PR #3, opened 2015-09-16
Slack bot setup

jessieay:
* Add `slack-bot-client` gem via Bundler
* Use `dotenv` to load ENV vars in development
* Add info on Slack API token to CONTRIBUTING.md


pkarman:
this looks sane to me, re: file layout, etc.
+1

jessieay:
@pkarman thanks for the review! 

Since there is going to be a DB and front end for this bot/app, it is going to turn into Rails real soon anyway, it turns out, but I appreciate the input on file organization!!

Issue #2, opened 2015-09-16
Front End Thoughts


melodykramer:
User should be able to:

1. Specify time/day message should be received
2. Specify message

Perhaps later, user should be able to:

1. See all messages sent to a specific user

jessieay:
@melodykramer when you say time/day, you mean for each *type* of message, or for each message?

In other words, my assumption was that we want a user to be able to set a rule like:

"On a new employee's first day of work, they should receive message ABC"
"On a new employee's second day of work, they should receive message XYZ"

We are going to run cron jobs once per day with `https://github.com/18F/cf-cron` so messages will always go out at the same time each day - does that work for you?

melodykramer:
Yes, that works!

melodykramer:
@jessieay One thing I thought about this morning - some users in Slack are not 18F people. So every 18F person should receive the message, but not a member of the public joining a single channel as a guest. As long as there's a way to differentiate 18F from non-18F, the method you described works perfectly.

jessieay:
Good thinking. I think a good v1 would be to have us just manually enter in
slack names and start dates into a DB since the Team API doesn't yet have
that info and it seems like that might take a while to setup and populate
so would rather not let it be a blocker.

Another thought I had is that we are going to need some kind of auth system
for this site if it is going to have a front end where people can update
messages. Is there a preferred auth system around here?

mbland:
Actually, it'd be damn easy to add it to the Team API. You can start putting the data in right away (a lot of Slack usernames are already in there), and then it should just show up automatically.

We host a number of sites (the Hub, Tock, pages-staging) behind an instance of oauth2_proxy that I administer. It's pretty easy to add a new site behind that interface, too.

jessieay:
Thanks for the input Mike.

As for adding fields to the team API, I will try to get on that now so it
doesn't block development of this bot.

Can you link me to some commits or repos re: oauth2. Also wondering if we
could just used simple auth (one username and pw for anyone to access the
site) for now since nothing in this app will require security, we just
don't want random people updating or deleting data. Thoughts?

mbland:
It's pretty easy to put a new site behind the oauth2_proxy by adding a new nginx config to the Hub. https://github.com/18F/hub/blob/master/deploy/etc/nginx/vhosts/tock.conf is the version for Tock that's currently running. (Note that checking a config into this dir isn't sufficient; we update the config on the server, then check it in to have a record of what's running.)

[...]

PR #10, opened 2015-09-18
User can create scheduled messages

jessieay:
* Body, day count are required fields
* Addresses front end requirements outlined in https://github.com/18F/dolores-landingham-bot/issues/2

screen shot 2015-09-18 at 11 04 22 am


gemfarmer:
@jessieay it might be good to have a message title or identifier (other than id). Other than that :+1:

jessieay:
@gemfarmer nice! I like that idea. Will add.

PR #13, opened 2015-09-18
Add service class to send bot messages

jessieay:
* Class invokes other classes to match employees with messages based on
  start date

screen shot 2015-09-18 at 2 52 00 pm


jessieay:
@gemfarmer

app/services/message_employee_matcher.rb
```diff
@@ -0,0 +1,17 @@
+class MessageEmployeeMatcher
+  def initialize(message)
+    @message = message
+  end
+
+  def run
+    Employee.where(started_on: day_count.days.ago).pluck(:id)
```
jessieay: might make more sense to return `slack_username` here since we are using the `id` to find the object/slack username right after this...

gemfarmer:
:+1:

jessieay:
@gemfarmer thanks for the review - merging :sheep:

PR #22, opened 2015-09-22
Add "add employee" to sidebar

jessieay:
* Also remove tabs from layout


gemfarmer:
Nice! Adding the default view as a sidebar option. Looks good on my end :+1:

PR #26, opened 2015-09-23
Add basic auth as stopgap

### LINKED ISSUE #25: Should there be a password?
Can anyone see the full message history? If so, I think there should be some kind of login or password, to limit this to 18F employees.

jessieay:
* Fixes #25


gemfarmer:
This works and can be added :+1:

PR #27, opened 2015-09-23
Set up DJ / Clockwork

### LINKED ISSUE #4: Run task daily that sends out appropriate bot messages
Using cloud foundry cron: https://github.com/18F/cf-cron

jessieay:
Fixes #4 
* Use Clockwork to send scheduled messages once per day at 3:30pm ET
* Use DJ to queue messages in an async manner


jessieay:
Don't have any automated tests for clockwork here since I felt like that would be testing the library itself...I did set clockwork up to send messages locally 2 minutes in the future and I confirmed that `MessageSender.new.delay.run` was triggered. Didn't confirm that messages actually went out via Slack because MessageSender has already been tested.

@gemfarmer do you think I should work to add some type of integration test for this new setup?

jessieay:
also interested in a review by @aidan since his foreman/clockwork setup on C2 was the inspiration for this approach :stars:

gemfarmer:
@jessieay When I tested it locally I received a `Sending scheduled messages` on the server, similarly confirming that `MessageSender.new.delay.run` was triggered, but no message was actually sent to my slack. So that interaction could definitely be tested.

Because tests are in place for `MessageSender.new.delay.run` without clockwork, I think you are right that there is no need for additional testing on the cron job itself.

config/clock.rb
```diff
@@ -0,0 +1,10 @@
+require "clockwork"
+require_relative "boot"
+require_relative "environment"
+
+module Clockwork
+  every(1.day, "scheduled_messages.send", at: "19:30", tz: "UTC") do
+    puts "Sending scheduled messages"
+    MessageSender.new.delay.run
```
gemfarmer: @jessieay This should be `MessageSender.new.run`. With the `delay` method, the test passes, but the message doesn't actually send.

jessieay: Hey! OMG I just realized why -we need to add a worker process to the Procfile. SO glad you checked this out. I will try to write an integration test for it

jessieay: I am running it with delayed job so that the jobs all get saved into the DB and re-run if they fail. So still want `delay` but do need to update setup

gemfarmer:
There were a few logic gaps that I found that were causing the MessageSender to error out.
In `slack_channel_id_finder.rb` add check for `slack_user`

```ruby
if slack_user
      slack_user_id = slack_user["id"] 
      chat = client.im_open(user: slack_user_id)
      chat["channel"]["id"]
end
```
In `message_sender.rb` add something like the following, so that the job doesn't get stuck on one faulty slack user:
```ruby
if !channel_id
  next
else
  client.chat_postMessage(
    channel: channel_id,
    as_user: true,
    text: message.body
  )
end
```

jessieay:
@gemfarmer Great points about rescuing errors in `MessageSender` - love the suggested code, too. My only hesitation is that we don't want to hide exceptions entirely...as they say, "exceptions should be exceptional" so I'd like to think of a way we can log them somehow rather than just continuing on. At the same time, I agree that we don't want everything to break when one slack username is wrong. Perhaps I will implement your suggestions now and then we can work on some type of error logging system in the near future...

jessieay:
@gemfarmer I'm going to update this after my other PR gets merged, since I think I will take advantage of the `FakeSlackAPI` in there to write tests for failure cases that might arise

aidan:
@jessieay i think you meant another Aidan, there are many like me but I'm not at 18F. Although you seem like you do interesting stuff, so keep up the good work!

jessieay:
@aidan ah yes, I meant "Aidan Feldman" - apologies about that, and thanks for stopping by our repo :rocket:

gemfarmer:
@jessieay I agree with your assessment. Patch it now, and let's plan on incorporating error logging in a new issue (#40)

config/clock.rb
afeld: Yep, this looks good.

afeld:
Didn't test locally but looks good overall! And yeah, we aren't doing testing of the clock either. :shrug:

gemfarmer:
Sounds like :ship:

jessieay:
merged at 6201e0f692751eb33d29f1c3c1df04a9d8c0fee6

nil cases taken care of :)

PR #29, opened 2015-09-23
Using omniauth from c2 project as template

### LINKED ISSUE #8: Auth
We need some kind of auth for this app so people don't start trolling us via Dolores.

See Mike's suggestion here: https://github.com/18F/dolores-landingham-bot/issues/2#issuecomment-141159857

I would also be OK with basic Rails authentication for this alpha version: http://api.rubyonrails.org/classes/ActionController/HttpAuthentication/Basic.html -- but we should double check that there wouldn't be a problem with us using that due to this being a gov website.

### LINKED ISSUE #31: Make Application Publicly Accessible
The app created through MyUSA is currently privately accessible only to developers on this project. We need to make it "Public", then screen only for gsa.gov email addresses.

gemfarmer:
Work in progress. Fixes #8 and Fixes #31


gemfarmer:
@jessieay Should be good to "merge"

Gemfile
```diff
@@ -11,6 +11,7 @@ gem "simple_form"
 gem "slack-ruby-client"
 gem "uglifier"
 gem "sass-rails"
+gem 'omniauth-myusa', git: 'https://github.com/18F/omniauth-myusa.git'
```
jessieay: dub quotes!

Gemfile
```diff
@@ -33,4 +34,5 @@ end
 
 group :staging, :production do
   gem "rack-timeout"
-end
+  gem 'rails_12factor'
```
jessieay: `'` -> `"`

Gemfile
jessieay: also, I believe you can do `github` here instead and it will read like

`gem "omniauth-myusa", github: "18F/omniauth-myusa"`

app/controllers/application_controller.rb
```diff
@@ -1,14 +1,22 @@
 class ApplicationController < ActionController::Base
-  protect_from_forgery with: :exception
-  before_filter :authenticate
 
-  private
+	protect_from_forgery with: :exception
```
jessieay: 2 space indentation, please

gemfarmer:
@jessieay Syntax changes made. Does it work for you locally?

jessieay:
@gemfarmer locally I see the link to log in but I am not being prevented from seeing the site before logging in - is that the desired behavior?

gemfarmer:
Hmm. No that is not. @jessieay could you clear your cookies and try again when you get a chance?

gemfarmer:
*if you didn't clear your cookies already, that is. If you do that and also try in incognito mode and you are getting the same result, we might need to redo something locally

jessieay:
@gemfarmer yes, saw the same in incognito. Do I need to be on a specific port or something?

gemfarmer:
Hmm. I have localhost:5000 listed in the app settings. Are you using localhost:5000 or 127.0.0.1:5000?

gemfarmer:
@jessieay let me know if you aren't getting the MyUSA auth on page load in `localhost:5000`. If that passes, then I'm going to push to master.

jessieay:
Just authed at port 5000 - went all the way through but once I entered in my password I got this: 

screen shot 2015-09-25 at 11 01 21 am

jessieay:
WORKS :shipit:

PR #39, opened 2015-09-25
Fixes #30:  Escapes all special characters to their HTML entities.

### LINKED ISSUE #30: Format messages in case of special characters
https://api.slack.com/docs/formatting

ccostino:
This will escape all special characters to their corresponding HTML entities, but it will also display the messages normally when viewing/editing them.

My Rails-fu is not as great as my Django-fu, so if there is a better way of accomplishing this please let me know!

### DIFF
--- a/app/models/scheduled_message.rb
+++ b/app/models/scheduled_message.rb
@@ -3,15 +3,17 @@ class ScheduledMessage < ActiveRecord::Base
   validates :days_after_start, presence: true
   validates :title, presence: true
 
-  before_save :escape_special_characters
+  before_save :escape_specific_characters
 
   def body
-    self[:body].html_safe
+    self[:body].html_safe if self[:body]
   end
 
   protected
 
-  def escape_special_characters
-    self.body = ERB::Util::html_escape(self[:body])
+  def escape_specific_characters
+    self.body = self[:body].gsub('&', '&amp;')
+    self.body = self[:body].gsub('<', '&lt;')
+    self.body = self[:body].gsub('>', '&gt;')
   end
 end
--- a/app/models/scheduled_message.rb
+++ b/app/models/scheduled_message.rb
@@ -2,18 +2,4 @@ class ScheduledMessage < ActiveRecord::Base
   validates :body, presence: true
   validates :days_after_start, presence: true
   validates :title, presence: true
-
-  before_save :escape_specific_characters
-
-  def body
-    self[:body].html_safe if self[:body]
-  end
-
-  protected
-
-  def escape_specific_characters
-    self.body = self[:body].gsub('&', '&amp;')
-    self.body = self[:body].gsub('<', '&lt;')
-    self.body = self[:body].gsub('>', '&gt;')
-  end
 end
--- a/app/models/scheduled_message.rb
+++ b/app/models/scheduled_message.rb
@@ -2,4 +2,16 @@ class ScheduledMessage < ActiveRecord::Base
   validates :body, presence: true
   validates :days_after_start, presence: true
   validates :title, presence: true
+
+  before_save :escape_special_characters
+
+  def body
+    self[:body].html_safe
+  end
+
+  protected
+
+  def escape_special_characters
+    self.body = ERB::Util::html_escape(self[:body])
+  end
 end
--- a/app/services/message_formatter.rb
+++ b/app/services/message_formatter.rb
@@ -4,7 +4,7 @@ class MessageFormatter
   end
 
   def escape_slack_characters
-    escaped_body = message_body
+    escaped_body = message.body
     escaped_body = escaped_body.gsub('&', '&amp;')
     escaped_body = escaped_body.gsub('<', '&lt;')
     escaped_body = escaped_body.gsub('>', '&gt;')
@@ -14,8 +14,4 @@ class MessageFormatter
   private
 
   attr_reader :message
-
-  def message_body
-    message.body || ""
-  end
 end
--- a/app/services/message_formatter.rb
+++ b/app/services/message_formatter.rb
@@ -1,0 +1,21 @@
+class MessageFormatter
+  def initialize(message)
+    @message = message
+  end
+
+  def escape_slack_characters
+    escaped_body = message_body
+    escaped_body = escaped_body.gsub('&', '&amp;')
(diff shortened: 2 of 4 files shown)


jessieay:
@ccostino I would send a test slack message via the bot to see what happens with this change. I am not sure if it makes sense to escape all special characters (eg: if Slack will re-encode), per your comment in #30

ccostino:
Sure enough, just the three characters need to be escaped, nothing else.  The cool thing is any formatting you include will show through in the message properly too, so if you add the *, _, `, etc., to the message, when it is sent through Slack it appears nicely to the user.

gemfarmer:
That's awesome. The people at slack are some swell folks, aren't they? :sweat_smile:

ccostino:
I'd say so! :smile:

app/models/scheduled_message.rb
```diff
@@ -2,4 +2,18 @@ class ScheduledMessage < ActiveRecord::Base
   validates :body, presence: true
   validates :days_after_start, presence: true
   validates :title, presence: true
+
+  before_save :escape_specific_characters
+
+  def body
+    self[:body].html_safe if self[:body]
```
jessieay: are these `self`s necessary? 

also, it feels weird to me to re-define `body` since that is an attribute on this model. How about a method name like `html_safe_body`?

I tend to prefer this style when checking for presence:

```ruby
body && body.html_safe
```

but since we are validating the presence of body (above), shouldn't there always be one?

app/models/scheduled_message.rb
```diff
@@ -2,4 +2,18 @@ class ScheduledMessage < ActiveRecord::Base
...
+  end
+
+  protected
+
+  def escape_specific_characters
```
jessieay: we aren't using this anywhere...

app/models/scheduled_message.rb
```diff
@@ -2,4 +2,18 @@ class ScheduledMessage < ActiveRecord::Base
   validates :body, presence: true
   validates :days_after_start, presence: true
   validates :title, presence: true
+
+  before_save :escape_specific_characters
```
jessieay: What about encoding on send (in the service class) rather than in a callback? I try to eliminate ALL callbacks in Rails when possible. Good blog post for ref: http://samuelmullen.com/2013/05/the-problem-with-rails-callbacks/

app/models/scheduled_message.rb
jessieay: oops, I was totally wrong about that. We are using it in the callback above (which I am suggesting we get rid of and instead make this a public method that we use in the `MessageSender` service class...thoughts?)

app/models/scheduled_message.rb
ccostino: Totally fine on the style change, but in terms of validating the presence, this becomes an issue when trying to enter a new message.  There is no body yet on the object since it is being created at that point, so there is nothing to call html_safe on.  Since this method is overriding just accessing the property directly, it cannot assume that is instantiated in that case.

app/models/scheduled_message.rb
ccostino: Ah, this is great, thank you!  I will shift some of this around to improve. :-)

app/services/message_sender.rb
```diff
@@ -15,12 +16,19 @@ def run
...
           channel: channel_id,
           as_user: true,
-          text: message.body
+          text: escaped_message_body
         )
       end
     end
   end
 
+  def escape_specific_characters(message_body)
```
jessieay: since we're only using this method inside this class, I think it can be `private`

jessieay: this might be kind of overkill...but what would you think about making a baby service class for this? That way, it would be easy to unit test

ccostino: Are you thinking of a small utils-like class for the Message/MessageSender object(s) to deal with any content formatting or manipulation instead?  This way if we have to add other things of this nature in the future they have a dedicated place to live?

jessieay: Yeah, likely a class in `app/services` that takes a message body as an argument and returns the message encoded. Would basically just include the logic in this method, and would allow us to write some AWESOME unit tests (otherwise, we'll have to test via `MessageSender` which seems kinda messy)

ccostino:
Interesting, I'm not sure why Git thinks there is still a conflict despite me merging everything from master before adding the latest changes.  It could be because I removed an extraneous comma during the initial merge itself.  I tried to account for that but no dice.  For whatever reason line 65 in app/services/message_sender.rb is the point of contention.

jessieay:
Carlo, it looks like you haven't rebased against master, since there are a lot of changes on this branch that were done on master. Can you try again?

ccostino:
Yeah let me redo this - I tried rebasing last time and didn't like me pushing in that case but I'm thinking I just need to force it.

ccostino:
There we go, that's a lot better. :-)

jessieay:
@ccostino yeah, one downside of rebasing is that you will always need to force push (which I would never advise on master, but am OK with on branches)

app/services/message_formatter.rb
```diff
@@ -0,0 +1,21 @@
...
+    escaped_body = escaped_body.gsub('>', '&gt;')
+    escaped_body
+  end
+
+  private
+
+  attr_reader :message
+
+  def message_body
+    message.body || ""
```
jessieay: since there is a presence validation in `message.body`, why do we have this?

spec/services/message_formatter_spec.rb
```diff
@@ -0,0 +1,17 @@
+require "rails_helper"
+
+describe MessageFormatter do
```
jessieay: nice test!

app/services/message_formatter.rb
ccostino: Indeed there is!  It's a holdover from what I had done earlier and not accounting for that, let's remove that since it's not needed.

spec/services/message_formatter_spec.rb
ccostino: Thanks, modeled after one of yours - the factory setup made it easy! :-)

jessieay:
Looks good to merge! :shipit:

ccostino:
Code has been merged!

PR #48, opened 2015-09-26
Remove unnecessary bullet initializer

### LINKED ISSUE #43: Bullet initializer?
Seems like a Bullet initializer made it in somehow - do we want to include that in every environment? Seems like a tool we should use locally to find N+1 queries but perhaps not needed to run on prod. Thoughts?

https://github.com/flyerhzm/bullet#configuration

jessieay:
* Fixes #43


ccostino:
If this wasn't used at all then I'd say it's safe to remove this!

jessieay:
@ccostino Thanks for the review! merged

PR #50, opened 2015-10-01
Add ability to test Scheduled Messages

### LINKED ISSUE #45: Provide ability to test messages
As someone who is configuring the Dolores Landhingham bot, I would like the ability to test the message I am creating so that I can see that it is formatted correctly in Slack itself.

We could provide a test link next to the edit link in the messages list, and there may be a couple of ways of accomplishing this:

* If you're logged into the app, having it automatically send them to just you as you test.
* Providing the ability to specify a person(s) who would also like to test/preview the messages.

jessieay:
* Enter slack username to test a message

Fixed #45 

https://trello.com/c/RafIoY6J

### DIFF
--- a/app/controllers/test_messages_controller.rb
+++ b/app/controllers/test_messages_controller.rb
@@ -1,0 +1,25 @@
+class TestMessagesController < ApplicationController
+  def new
+    @message = ScheduledMessage.find(params[:scheduled_message_id])
+  end
+
+  def create
+    message = ScheduledMessage.find(params[:scheduled_message_id])
+    employee = Employee.find_by(slack_username: slack_username)
+
+    if employee
+      MessageSender.new(employee, message).delay.run
+      flash[:notice] = "Test message sent successfully"
+      redirect_to scheduled_messages_path
+    else
+      flash[:error] = "Could not update employee with the Slack username you entered. Please make sure there is an employee record for the username listed in #{employees_path}"
+      redirect_to scheduled_messages_path
+    end
+  end
+
+  private
+
+  def slack_username
+    params[:test_message][:slack_username]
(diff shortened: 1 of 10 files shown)


app/controllers/test_messages_controller.rb
```diff
@@ -0,0 +1,25 @@
...
+  def new
+    @message = ScheduledMessage.find(params[:scheduled_message_id])
+  end
+
+  def create
+    message = ScheduledMessage.find(params[:scheduled_message_id])
+    employee = Employee.find_by(slack_username: slack_username)
+
+    if employee
+      MessageSender.new(employee, message).delay.run
```
gemfarmer: I think this should just be `MessageSender.new(employee, message).run`? When I tested it, it only works when I remove delay.

gemfarmer:
@jessieay other than my one line note, this looks good to merge!

app/controllers/test_messages_controller.rb
jessieay: We want these to run async (with Delayed Job) so the request doesn't time out waiting for the messages to send (this might happen if the Slack API is down or being slow)

I also saw this wasn't working yesterday...and the error message is "not_authed" which I am pretty sure means DJ isn't seeing the environment variables that allow it to auth with Slack (the API key). Will look into this more now.

jessieay: YES just got it to work - realized the reason it wasn't authing was that we were configuring the Slack client in `initialize`...so when we called `delay`, the Slack client wasn't working because there was a disconnect between the `initialize` and the `run` - moved `configure_slack` into `run` and now IT WORKS

Issue #33, opened 2015-09-24
Validation on slack username


jessieay:
when someone enters a slack username for an employee record, we should validate that it does not have an `@` at the beginning

gemfarmer:
Slightly more complicated, but this should also validate that that slack user exists. On top of that, and this might be a new issue, we should store the slack user id so that if they change their slack "name", it doesn't throw off our reference.

ccostino:
Is there a specific format that a Slack username must be, e.g., only letters, alpha-numeric, etc.?  I've been trying to sift through the documentation looking for an answer.

ccostino:
I found it right in the profile settings:  "Usernames must be lowercase and cannot contain spaces."  So it looks like lowercase alpha only.

jessieay:
Closing in favor of https://trello.com/c/MCGArovt

jessieay:
Via Carlo:

Upon further research it's not just lowercase letters as I see folks with underscores, numbers, etc., it's just that the letters have to be (or are just converted automatically) to lowercase. I still haven't been able to pin-point documentation that specifies what characters are specifically allowed, I'll keep searching for a bit.

ccostino:
Looks like to validate a username it should just be a call for looking up user info and checking for the user_not_found error: https://api.slack.com/methods/users.info

ccostino:
Okay, there's a simpler solution as we already retrieve the list of users and search through that - I'm going to leverage that functionality and build from there!

ccostino:
@gemfarmer, I think the storing of the Slack user ID should be separate from this and may not necessarily be needed for this particular use case (though I think it's still a good idea anyway).

jessieay:
Merged at https://github.com/18F/dolores-landingham-bot/commit/823871cd04e72db5d76aa7f1dfa78a067146aa9d

PR #63, opened 2015-10-03
Issue #33:  Slack Username Validation

ccostino:
This is the initial changeset with full support and passing tests for the Slack username validation.  There is still one outstanding thing to investigate further with the employee model validation tests as the "allow_values" shoulda matcher method seems to be failing.  I want to look into that a bit further to see if it's something that needs to be reported to the shoulda matchers (and gather full details if so), but otherwise this is ready for review and discussion!  Thank you @jessieay for your help with this and coming up with a cleaner approach, especially with the tests!

### DIFF
--- a/app/controllers/employees_controller.rb
+++ b/app/controllers/employees_controller.rb
@@ -48,6 +48,6 @@ class EmployeesController < ApplicationController
   end
 
   def unknown_employee(slack_username)
-    !EmployeeFinder.new.employee_exists?(slack_username)
+    !EmployeeFinder.new(slack_username).existing_employee?
   end
 end
--- a/app/controllers/employees_controller.rb
+++ b/app/controllers/employees_controller.rb
@@ -6,13 +6,10 @@ class EmployeesController < ApplicationController
   def create
     @employee = Employee.new(employee_params)
 
-    unless EmployeeFinder.new.employee_exists?(@employee.slack_username)
-      flash.now[:error] = "\"#{@employee.slack_username}\" is not a valid employee (that username was not found)."
+    if unknown_employee(@employee.slack_username)
+      flash.now[:error] = "There is not a slack user with the username \"#{@employee.slack_username}\" in your organization."
       render action: :new
-      return
-    end
-
-    if @employee.save
+    elsif @employee.save
       flash[:notice] = "Thanks for adding #{@employee.slack_username}"
       redirect_to root_path
     else
@@ -32,13 +29,10 @@ class EmployeesController < ApplicationController
   def update
     @employee = Employee.find(params[:id])
 
-    unless EmployeeFinder.new.employee_exists?(@employee.slack_username)
-      flash.now[:error] = "\"#{@employee.slack_username}\" is not a valid employee (that username was not found)."
+    if unknown_employee(params[:employee][:slack_username])
+      flash.now[:error] = "There is not a slack user with the username \"#{params[:employee][:slack_username]}\" in your organization."
       render action: :edit
-      return
-    end
-
-    if @employee.update(employee_params)
+    elsif @employee.update(employee_params)
       flash[:notice] = "Employee updated successfully"
       redirect_to employees_path
     else
@@ -51,5 +45,9 @@ class EmployeesController < ApplicationController
 
   def employee_params
     params.require(:employee).permit(:slack_username, :started_on)
+  end
+
+  def unknown_employee(slack_username)
+    !EmployeeFinder.new.employee_exists?(slack_username)
   end
 end
--- a/app/controllers/employees_controller.rb
+++ b/app/controllers/employees_controller.rb
@@ -6,6 +6,12 @@ class EmployeesController < ApplicationController
   def create
     @employee = Employee.new(employee_params)
 
+    unless EmployeeFinder.new.employee_exists?(@employee.slack_username)
+      flash.now[:error] = "\"#{@employee.slack_username}\" is not a valid employee (that username was not found)."
+      render action: :new
+      return
+    end
+
     if @employee.save
       flash[:notice] = "Thanks for adding #{@employee.slack_username}"
       redirect_to root_path
@@ -25,6 +31,12 @@ class EmployeesController < ApplicationController
 
   def update
     @employee = Employee.find(params[:id])
+
+    unless EmployeeFinder.new.employee_exists?(@employee.slack_username)
+      flash.now[:error] = "\"#{@employee.slack_username}\" is not a valid employee (that username was not found)."
+      render action: :edit
+      return
+    end
 
     if @employee.update(employee_params)
       flash[:notice] = "Employee updated successfully"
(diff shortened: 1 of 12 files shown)


app/controllers/employees_controller.rb
```diff
@@ -6,6 +6,12 @@ def new
   def create
     @employee = Employee.new(employee_params)
 
+    unless EmployeeFinder.new.employee_exists?(@employee.slack_username)
```
jessieay: would it be possible to change this to an `if` instead of an `unless`? I find `unless` statements hard to parse because there's a negative - easier to say "if this, then that" rather than "unless not this, then that" (as I write this, I am not sure if that is even correct...more evidence that unless is hard!)

Also, since we are using the same statement on this line and L35, perhaps the logic can be extracted into a private method with a nice name....

app/controllers/employees_controller.rb
```diff
@@ -6,6 +6,12 @@ def new
...
+      flash.now[:error] = "\"#{@employee.slack_username}\" is not a valid employee (that username was not found)."
+      render action: :new
+      return
```
jessieay: don't need an explicit return in ruby :)

app/controllers/employees_controller.rb
```diff
@@ -6,6 +6,12 @@ def new
   def create
     @employee = Employee.new(employee_params)
 
+    unless EmployeeFinder.new.employee_exists?(@employee.slack_username)
+      flash.now[:error] = "\"#{@employee.slack_username}\" is not a valid employee (that username was not found)."
```
jessieay: how about just `"There is not a slack user with the username \"#{@employee.slack_username}\" in your organization"`

spec/models/employee_spec.rb
```diff
@@ -5,6 +5,11 @@
      subject { create(:employee) }
      it { should validate_presence_of(:slack_username) }
      it { should validate_uniqueness_of(:slack_username) }
+     # These fail - why?
+     #it { should allow_values("test_user_1", "testuser2", "test.user.3").for(:slack_username) }
+     #it { should_not allow_values("TEST USER 1", "test_user_2?", "test'user'3").for(:slack_username) }
```
jessieay: remove commented code?

app/services/employee_finder.rb
```diff
@@ -0,0 +1,23 @@
+require "slack-ruby-client"
+
+class EmployeeFinder
+  def initialize
+    configure_slack
+  end
+
+  def employee_exists?(slack_username)
```
jessieay: so the way we've been writing these services classes, they all use the `run` method to process logic...thoughts on following that pattern here as well?

spec/services/slack_user_finder_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    it "returns false if a user was not found" do
+      fake_slack_user_name = "testusernamethatdoesnotexist"
+
+      user_exists = SlackUserFinder.new(
+        fake_slack_user_name,
+        Slack::Web::Client.new
+      ).user_exists?
+
+      expect(user_exists).to be false
```
jessieay: another way to write this, if we keep the method name as a boolean, would be:


```ruby
user = SlackUserFinder.new(fake_slack_user_name, Slack::Web::Client.new)

expect(user).to be_user_exists
```
one thing I like about doing that is it reveals the quality of boolean method names - now that I wrote it this way, it seems the name of this method is repetitive / off. If we keep it a boolean, perhaps something like `is_in_organization?` would be better...

Source :https://www.relishapp.com/rspec/rspec-expectations/v/2-2/docs/matchers/predicate-matchers#matcher-arguments-are-passed-on-to-the-predicate-method

app/controllers/employees_controller.rb
ccostino: In this case we do though, without it the controller action was continuing to flow, causing it to add the invalid employee anyway!  Unless (pun not intended, haha!) there's another/better way to immediately stop and return?

app/controllers/employees_controller.rb
ccostino: More concise, I like it!

app/controllers/employees_controller.rb
jessieay: we could do an `if` and `elsif` rather than two `if`s!

app/services/employee_finder.rb
ccostino: I'm definitely a fan of consistency, but also explicit naming conventions and the purpose of this method isn't to process anything but instead to find something.  If these methods are to be treated as the main entry points for the service classes regardless of what they do, then perhaps a more generic name might be better?

spec/models/employee_spec.rb
ccostino: Yes, once I uncover what's going on with these. :-)

app/controllers/employees_controller.rb
ccostino: Yes, definitely on both accounts.  I need to ban `unless` from my Ruby vocabulary... :-)

spec/services/slack_user_finder_spec.rb
ccostino: Oh interesting, thank you for the heads up with the RSpec info (something I'm still diving into).  As for the name, are you thinking that it should be changed in the `EmployeeFinder` service class?  I'm not sure about `is_in_organization?` only because that sounds like it should accept an argument of an organization to me and we're just checking for the existence of a user in the organization we're a part of when logged in.  I agree that the naming could be improved though, especially in light of how it flows with the Rspec matchers.

app/controllers/employees_controller.rb
ccostino: Oh, right!

spec/models/employee_spec.rb
ccostino: In thinking about this a bit more, while this would be an interesting investigation to go down I don't think it's necessary at this time for the sake of finishing this PR and moving forward to other things.  If it rears it's head again we can revisit it then.

spec/services/employee_finder_spec.rb
```diff
@@ -0,0 +1,35 @@
+require "rails_helper"
+
+describe EmployeeFinder do
+  describe "#employee_exists?" do
+    it "returns true if an employee exists in Slack" do
```
jessieay: how about `it "returns true if an employee exists in the Slack organization associated with the auth token" do`

it's kinda wordy, but will be helpful for future people reading these tests!

jessieay:
One comment on a spec description - otherwise looks SO GREAT please please merge this is awesome

spec/services/employee_finder_spec.rb
ccostino: Oh definitely - much better to be explicit rather than implicit so I agree!  Consider it done. :-)

ccostino:
Great, thank you so much for looking things over and providing feedback @jessieay, I appreciate it and they've been very helpful!

ccostino:
Branch is now merged!

jessieay:
nice! for future: if you rebase your branch and force push before merging to master, the PR will automatically get closed when you merge (because the commit hashes will line up) :sparkles:

Issue #28, opened 2015-09-23
Add tagging support to scheduled messages


melodykramer:
Not for MVP:

Way for messages to be tagged so that people can sign up for specific types of messages: i.e. refresher on benefits (receive all benefits msgs.)

jessieay:
Closing in favor of https://trello.com/c/MrJuYar8

ccostino:
Are there any concerns with using the acts_as_taggable_on gem to support this?

https://github.com/mbleigh/acts-as-taggable-on

jessieay:
Nope, sounds good to me. 

I think there are 2 steps to this:

1) Allow people creating a scheduled message to tag the message
2) Create some kind of mechanism where people can subscribe to a tag. Would need to discuss further with someone (probably Melody/Andrew) about that because I am guessing we wouldn't want to spread out the messages in the same way for subscribers that we do for newbies. Perhaps a secondary ordering system for topic subscribers...would have to think further about this

ccostino:
Agreed - I wonder if we should split this out into two tickets?  Add support for the tagging itself in this one and then developing the mechanism for folks to subscribe in another?  That second step might require similar permissions work as with #53 and establishing preference management within the app itself.

jessieay:
@ccostino yeah, feel free to split!

jessieay:
@ccostino moving this GH issue back to next up - I see #66 is about allowing users to subscribe to specific tagged messages. What about just letting users view all messages with a specific tag first?

PR #67, opened 2015-10-08
Connects #28:  Initial support for tags on scheduled messages.

ccostino:
This changeset adds support for tags on scheduled messages via the acts_as_taggable_on gem.  New migrations are included and a couple of tests have been updated to account for the new functionality.

### DIFF
--- a/app/controllers/scheduled_messages_controller.rb
+++ b/app/controllers/scheduled_messages_controller.rb
@@ -38,6 +38,6 @@ class ScheduledMessagesController < ApplicationController
   private
 
   def scheduled_message_params
-    params.require(:scheduled_message).permit(:body, :days_after_start, :title, :tag_list)
+    params.require(:scheduled_message).permit(:body, :days_after_start, :tag_list, :title)
   end
 end
--- a/app/controllers/scheduled_messages_controller.rb
+++ b/app/controllers/scheduled_messages_controller.rb
@@ -38,6 +38,6 @@ class ScheduledMessagesController < ApplicationController
   private
 
   def scheduled_message_params
-    params.require(:scheduled_message).permit(:body, :days_after_start, :title)
+    params.require(:scheduled_message).permit(:body, :days_after_start, :title, :tag_list)
   end
 end
--- a/app/views/scheduled_messages/_form.html.erb
+++ b/app/views/scheduled_messages/_form.html.erb
@@ -1,8 +1,8 @@
 <%= simple_form_for @scheduled_message do |form| %>
   <%= form.input :title,
-    hint: "This is for identifying purposes only, the employee will not see the message title" %>
+    hint: "This is for identifying purposes only, the employee will not see the message title." %>
   <%= form.input :body, label: "Message body" %>
   <%= form.input :days_after_start, label: "Days after employee starts to send message" %>
-  <%= form.input :tag_list, label: "Tags (separated by commas)" %>
+  <%= form.input :tag_list, label: "Tags", hint: "Separate tags with spaces to enter multiple tags at once." %>
   <%= form.button :submit %>
 <% end %>
--- a/app/views/scheduled_messages/_form.html.erb
+++ b/app/views/scheduled_messages/_form.html.erb
@@ -3,6 +3,6 @@
     hint: "This is for identifying purposes only, the employee will not see the message title." %>
   <%= form.input :body, label: "Message body" %>
   <%= form.input :days_after_start, label: "Days after employee starts to send message" %>
-  <%= form.input :tag_list, label: "Tags", hint: "Separate tags with spaces to enter multiple tags at once." %>
+  <%= form.input :tag_list, label: "Tags", hint: "Separate tags with commas to enter multiple tags at once." %>
   <%= form.button :submit %>
 <% end %>
--- a/app/views/scheduled_messages/_form.html.erb
+++ b/app/views/scheduled_messages/_form.html.erb
@@ -3,6 +3,6 @@
     hint: "This is for identifying purposes only, the employee will not see the message title" %>
   <%= form.input :body, label: "Message body" %>
   <%= form.input :days_after_start, label: "Days after employee starts to send message" %>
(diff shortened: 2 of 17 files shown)


app/controllers/scheduled_messages_controller.rb
```diff
@@ -38,6 +38,6 @@ def update
   private
 
   def scheduled_message_params
-    params.require(:scheduled_message).permit(:body, :days_after_start, :title)
+    params.require(:scheduled_message).permit(:body, :days_after_start, :title, :tag_list)
```
jessieay: alpha order this list?

spec/features/create_scheduled_message_spec.rb
```diff
@@ -13,4 +13,18 @@
...
+
+  scenario "successfully with tags" do
+    login_with_oauth
+    visit root_path
+    click_on "Create scheduled message"
+
+    fill_in "Title", with: "Message title"
+    fill_in "Message body", with: "Message body"
+    fill_in "Days after employee starts to send message", with: 1
+    fill_in "Tags (seperated by commas)", with: "tag_one tag_two tag_three"
```
jessieay: "separated"

spec/features/edit_scheduled_messages_spec.rb
```diff
@@ -15,4 +15,24 @@
...
+    old_title = "Old title"
+    new_title = "New title"
+    tags = "tag_one, tag_two, tag_three"
+    create(:scheduled_message, title: old_title)
+
+    login_with_oauth
+    visit scheduled_messages_path
+    click_on "Edit"
+    fill_in "Title", with: new_title
+    fill_in "Tags (seperated by commas)", with: tags
```
jessieay: spelling!

spec/features/edit_scheduled_messages_spec.rb
```diff
@@ -15,4 +15,24 @@
     expect(page).to have_content "Scheduled message updated successfully"
     expect(page).to have_content new_title
   end
+
+  scenario "from list of scheduled messages with tags" do
```
jessieay: I might change this to something like `scenario "edits tags for a scheduled message" do`

spec/features/edit_scheduled_messages_spec.rb
jessieay: what happens in someone forgets to put commas between?

jessieay:
Should we require tags? This might be a melody question

spec/features/create_scheduled_message_spec.rb
ccostino: D'oh...  Thanks!

spec/features/edit_scheduled_messages_spec.rb
ccostino: The tags get added as a single string, which I'm not exactly comfortable with just leaving, so I am going to work on improving that a bit.

ccostino:
Definitely a @melodykramer question. :-)  Generally speaking I would lean toward no, but in the case of this application and the use-cases involved (with users subscribing to specific tags being the primary one), it might not make sense to not tag the messages.

app/controllers/scheduled_messages_controller.rb
ccostino: Yes!  I don't know why I missed that, I'm usually a stickler for such things myself.

spec/features/edit_scheduled_messages_spec.rb
ccostino: I will wordsmith this some more. :-)

melodykramer:
I would say yes to tags, if possible.

ccostino:
I will make sure tags are required, then!

app/views/scheduled_messages/_form.html.erb
```diff
@@ -1,7 +1,8 @@
 <%= simple_form_for @scheduled_message do |form| %>
   <%= form.input :title,
-    hint: "This is for identifying purposes only, the employee will not see the message title" %>
+    hint: "This is for identifying purposes only, the employee will not see the message title." %>
   <%= form.input :body, label: "Message body" %>
   <%= form.input :days_after_start, label: "Days after employee starts to send message" %>
+  <%= form.input :tag_list, label: "Tags", hint: "Separate tags with spaces to enter multiple tags at once." %>
```
jessieay: you changed this (from commas to spaces)! 

I think that comma separation is more typical in a case like this - what's the thinking behind spaces?

jessieay:
One comment on tag char separator - otherwise :ship: when ready!

app/views/scheduled_messages/_form.html.erb
ccostino: I did, and I was thinking ease of use in this case.  It's quicker/easier to just type a string of single words with spaces to add a bunch of tags.  There is a caveat though in that you then can't have multiple words make up a single tag.  I should ask before making that assumption:  if that is going to be a use-case then we would need to keep the comma (or some other delimiter) separation.

Furthermore, I did toy around with a small string substitution when not entering commas to put them in there (splitting the string on a space) in order to enforce multiple tags, but I wasn't a fan of it since that creates some "magic" behind the scenes (and now that it's being discussed, it makes the same assumption as noted above).  Obviously this would be okay if we explicitly states what happens, but let's nail down what behavior we're ultimately looking for.

@melodykramer: Can you please chime in with what you feel would be best?  Do you foresee adding tags that are made up of multiple words?  If so, is explicitly noting how tags should be separated (e.g., commas) in the help text enough for the app now (thinking MVP), or do you feel we would need a bit more support in the UI to prevent usage errors?  Thanks!

melodykramer:
I would say that there might be multiple words, to align this with our Slack channel names and guides elsewhere. I would recommend comma separated, if possible.

ccostino:
Okay, no worries (it simplifies the code a bit!) is it fine at this point to simply leave it as a note under the field to separate tags with commas and leave the rest up to the user to enter things correctly, or do we think we need more fine-tuned controls to ensure proper tag input/management?

ccostino:
This PR has been merged to master, thanks @jessieay and @melodykramer!

Issue #54, opened 2015-10-01
Add ability to delete employee record


jessieay:
Users should be able to delete records from the employee#index view. We should use a gem like paranoia to keep all records in the DB even if they are deleted https://github.com/radar/paranoia

jessieay:
Pr for this https://github.com/18F/dolores-landingham-bot/pull/71 -- didn't add paranoia...but maybe we should??

PR #71, opened 2015-10-16
[#54] Admin can delete employees

jessieay:
Connects #54

### DIFF
--- a/app/controllers/employees_controller.rb
+++ b/app/controllers/employees_controller.rb
@@ -41,6 +41,14 @@ class EmployeesController < ApplicationController
     end
   end
 
+  def destroy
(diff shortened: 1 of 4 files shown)


jessieay:
Connects #54

ccostino:
Looks good and works great, @jessieay!  Nice and simple. :-)  Regarding Waffle integration, if you put the phrase "Connects #54" into the description or commit message itself (description should work fine here since this is now in GitHub), that should tie the two cards together in Waffle.

This is good to go though, feel free to merge and deploy!

jessieay:
@ccostino thanks for the tip! works like a charm on waffle

Issue #75, opened 2015-10-16
Update formatting in messages#index


jessieay:
we show the whole message body now, which looks wacky when the message is long:

![screen shot 2015-10-16 at 11 32 24 am](https://cloud.githubusercontent.com/assets/601515/10550095/a773bcd8-73f9-11e5-8681-58bda6220164.png)

jessieay:
@melodykramer you agree that we should just show the first bit of the message on this page and allow someone to view the whole thing by clicking on the title?

melodykramer:
Yes, @jessieay - first bit is fine!

Issue #73, opened 2015-10-16
Improve flash message when test message send fails


jessieay:
Melody tried to send a test message to herself and the message failed. The current error message is:

`"Could not update employee with the Slack username you entered. Please make sure there is an employee record for the username listed in #{employees_path}"`

The error occurs when there is no employee record for the slack username entered into the test message box.

Seems like we could improve this error message to make the meaning more clear (I think we meant to use "find" instead of "update" anyway...that's a clear mistake). 

Suggestions on how to make this error message wording more clear?

@melodykramer

melodykramer:
Oops! Looks like that employee isn't in the system yet! Make sure you've entered the Slack handle for that employee.

Issue #51, opened 2015-10-01
Add flash notice/error styles


jessieay:
Right now, they are unstyled. Attaching screenshots as examples.

jessieay:
<img width="394" alt="screen shot 2015-10-01 at 3 31 19 pm" src="https://cloud.githubusercontent.com/assets/601515/10234022/02490f14-6860-11e5-9bf4-198d2c17b5aa.png">
![screen shot 2015-10-01 at 3 30 19 pm](https://cloud.githubusercontent.com/assets/601515/10234021/0248b58c-6860-11e5-8848-2e5d58902b40.png)

PR #77, opened 2015-10-16
[Connects #73, #51, #75] Fix small UI things

jessieay:
* Add styles for flash errors/notices
* Truncate message body in index view
* Improve error message when test send fails

Connects #73 
Connects #51 
Connects #75


jessieay:
Gonna self-merge since these are just little changes and are blocking usability (hard to use messages index when they are so weirdly formatted)

Issue #78, opened 2015-10-16
Default page should be messages index


jessieay:
Right now the root path is the new employee form. It makes more sense to show the list of messages, since that is the most interesting part of this app...I think?

ccostino:
I think this makes sense for right now.  In the future I could see a dashboard page of sorts that lets you easily get to whatever task you want to accomplish right away, but for right now the messages are definitely the highlight of the app. :-)

PR #79, opened 2015-10-16
Change root path the ScheduledMessages#index

jessieay:
* Connects #78


ccostino:
@jessieay:  I just checked this out quick and it's definitely good to go, thanks!

Issue #89, opened 2015-10-22
Fix tags not being loaded properly in the scheduled message edit form


ccostino:
I noticed an issue with the scheduled message form when loading a previously saved message.  The tags were not being displayed with the separator (a comma), causing them to be saved as a single tag upon updating the message.

PR #90, opened 2015-10-22
Connects #89:  Fixes an issue with the tags input in the scheduled messages edit form not preserving its separators.

ccostino:


gemfarmer:
looks good! merging

Issue #76, opened 2015-10-16
Allow time in addition to days after hire?


melodykramer:
We could make messages much shorter if we could post multiple times in a day. One way to do this is to say "What time would you like to post this?" 

Considerations: time zones, DST, Arizona (which has wacky DST rules)

ccostino:
@melodykramer and @jessieay:  I was thinking about this a bit more, and one solution I came up with to solve the disparity with time zones is that we could store that info with the rest of the employee info (along with their started on date) so that we could always send a message at their local time.  In other words we would be able to just set a time, e.g., 10:00 AM, on a scheduled message and the message would be sent at 10:00 AM local time for any given user.  This way folks on in the PST time zone wouldn't be getting their messages at 7:00 AM, for instance. :-)

Does this sound like a good approach?  It would appear that we'd need to add support for a timezone option to the employees, a time value for scheduled messages, and then do some enhancements to the message sender.

melodykramer:
This would be an excellent approach. (And useful for other things too.)

ccostino:
Great, thanks @melodykramer!

@jessieay and @gemfarmer:  I'm reading conflicting things about whether or not to set the default time zone in the application.  My take is that it reduces the probability of date/time related weirdness by not accounting for what the server's time is set.  If it's set in the app then it can be deployed anywhere without concern for this as we can expect it to behave consistently.

Do either of you have any thoughts or strong feelings about this?  I wanted to check before just setting this in the app.

melodykramer:
One thing I was thinking about this morning - I don't want these notes to show up on the weekends. And it's going to get REALLY hard to determine holiday Mondays . I'm not sure how to fix this, but thought I would bring it up in this thread. I think it's much harder to think "Oh, that's for day 104" than to think "That's for the 4th Friday in their second Month."

ccostino:
@melodykramer:  Hmmm, really good point about that.  On the one hand, any messages sent while you're offline will be there when you login again, but I can imagine it would be easy to miss those messages as well.

I'd tackle this in a separate issue which I can create, but thank you for bringing this up, it is something worth keeping in mind!

PR #93, opened 2015-10-29
Connects #76: Add support for time zones for employees and set times for messages

ccostino:
This changeset adds support for specific times to be set for scheduled messages and for setting time zones for employees.  The combination of these two things now allows an administrator to set scheduled messages to go out at specific times, e.g., 10:00 AM, and those messages well be sent respective to each employee's local time (meaning the message will be sent to an EST employee at 10 AM EST, and to a PST employee at 10 AM PST).

### DIFF
--- a/app/services/message_employee_matcher.rb
+++ b/app/services/message_employee_matcher.rb
@@ -4,14 +4,30 @@ class MessageEmployeeMatcher
   end
 
   def run
-    Employee.where(started_on: day_count.days.ago)
+    retrieve_matching_employees
   end
 
   private
 
   attr_reader :message
 
+  def retrieve_matching_employees
+    Employee.where(started_on: day_count.days.ago)
+      .select { |employee| match_time(employee.time_zone) && message_not_already_sent?(employee) }
+  end
+
   def day_count
     message.days_after_start
+  end
+
+  def match_time(time_zone)
+    employee_current_hour = Time.current.in_time_zone(time_zone).hour
+    employee_current_minute = Time.current.in_time_zone(time_zone).min
+
+    employee_current_hour == message.time_of_day.hour && employee_current_minute == message.time_of_day.min
+  end
+
+  def message_not_already_sent?(employee)
+    SentScheduledMessage.where(employee: employee, scheduled_message: message).count == 0
   end
 end
--- a/app/views/scheduled_messages/_form.html.erb
+++ b/app/views/scheduled_messages/_form.html.erb
@@ -5,6 +5,7 @@
     hint: "Don't forget, you can format your message using #{link_to "Slack's message formatting rules",
     "https://slack.zendesk.com/hc/en-us/articles/202288908-Formatting-your-messages", target: "_blank"}.".html_safe %>
   <%= form.input :days_after_start, label: "Days after employee starts to send message" %>
+  <%= form.input :time_of_day, label: "Time of day to send message (uses employee's local time zone)", ampm: true, input_html: {class: "date-time-select"} %>
   <%= form.input :tag_list, label: "Tags", hint: "Separate tags with commas to enter multiple tags at once.", input_html: {value: @scheduled_message.tag_list.to_s} %>
   <%= form.button :submit %>
 <% end %>
--- a/db/migrate/20151021194020_add_time_of_day_to_scheduled_messages.rb
+++ b/db/migrate/20151021194020_add_time_of_day_to_scheduled_messages.rb
@@ -1,0 +1,5 @@
+class AddTimeOfDayToScheduledMessages < ActiveRecord::Migration
+  def change
+    add_column :scheduled_messages, :time_of_day, :time, null: false, default: "12:00:00"
+  end
+end
--- a/spec/services/message_employee_matcher_spec.rb
+++ b/spec/services/message_employee_matcher_spec.rb
@@ -3,17 +3,49 @@ require "rails_helper"
 describe MessageEmployeeMatcher do
   describe "#run" do
     it "matches messages to users based on days after start" do
+      Timecop.freeze(Time.current)
+
       days_after_start = 3
-      scheduled_message = create(:scheduled_message, days_after_start: days_after_start)
-      _other_message = create(:scheduled_message, days_after_start: 2)
-      employee_for_scheduled_message = create(:employee, started_on: days_after_start.days.ago)
-      _other_employee = create(:employee, started_on: Date.today)
+      scheduled_message = create(:scheduled_message, days_after_start: days_after_start, time_of_day: Time.current.in_time_zone("Eastern Time (US & Canada)"))
+      employee = create(:employee, started_on: days_after_start.days.ago)
 
       matched_employees_and_messages = MessageEmployeeMatcher.new(scheduled_message).run
 
       expect(matched_employees_and_messages).to eq(
-          [ employee_for_scheduled_message ]
+          [ employee ]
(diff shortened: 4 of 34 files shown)


spec/services/message_employee_matcher_spec.rb
```diff
@@ -2,10 +2,18 @@
 
 describe MessageEmployeeMatcher do
   describe "#run" do
+    before do
```
jessieay: I prefer inlining `before` and `after`  actions to avoid the mystery guest problem - it's more duplication, but gives more transparency about the behavior of each spec. What do you think about doing that in this file?

app/services/message_employee_matcher.rb
```diff
@@ -14,4 +14,15 @@ def run
...
   end
+
+  def match_time(time_zone)
+    employee_current_hour = Time.current.in_time_zone(time_zone).hour
+    employee_current_minute = Time.current.in_time_zone(time_zone).min
+
+    employee_current_hour == message.time_of_day.hour && employee_current_minute == message.time_of_day.min
+  end
+
+  def message_already_sent?(employee)
```
jessieay: this is awesome - do you think we should add a spec for this service class that tests the logic in this method? (not directly, since it's private, but a test where the setup includes an already sent message)

app/services/message_employee_matcher.rb
```diff
@@ -4,7 +4,7 @@ def initialize(message)
   end
 
   def run
-    Employee.where(started_on: day_count.days.ago)
+    Employee.where(started_on: day_count.days.ago).select { |employee| match_time(employee.time_zone) && !message_already_sent?(employee) }
```
jessieay: there's a lot going on / to understand on this line - what about putting `Employee.where(started_on: day_count.days.ago).select { |employee| match_time(employee.time_zone)` into a small well-named method (similar to what you did with `message_already_sent?`) so it is easier to read and understand in smaller parts?

app/services/message_employee_matcher.rb
jessieay: also, what about changing this to `message_not_already_sent?` -- the `!` works, too, but since we're only using this method for one purpose, it might be nice to just have it return what we want rather than the opposite (am I making sense? it's early here in PST...)

app/views/scheduled_messages/_form.html.erb
```diff
@@ -5,6 +5,7 @@
     hint: "Don't forget, you can format your message using #{link_to "Slack's message formatting rules",
     "https://slack.zendesk.com/hc/en-us/articles/202288908-Formatting-your-messages", target: "_blank"}.".html_safe %>
   <%= form.input :days_after_start, label: "Days after employee starts to send message" %>
+  <%= form.input :time_of_day, label: "Time of day to send message (respects employee's local time zone)", ampm: true, input_html: {class: "date-time-select"} %>
```
jessieay: this is great - love the clear copy. instead of "respects" what about "uses"?

db/migrate/20151021194020_add_time_of_day_to_scheduled_messages.rb
```diff
@@ -0,0 +1,5 @@
+class AddTimeOfDayToScheduledMessages < ActiveRecord::Migration
+  def change
+    add_column :scheduled_messages, :time_of_day, :time, null: false, default: "12:00:00"
```
jessieay: is this default for UTC?

spec/services/message_sender_spec.rb
```diff
@@ -1,6 +1,14 @@
 require "rails_helper"
 
 describe MessageSender do
+  before do
```
jessieay: same comment re: inlining

spec/services/message_employee_matcher_spec.rb
ccostino: Do you mean just taking them out and copying whatever they do into each spec that utilizes them directly?

app/services/message_employee_matcher.rb
ccostino: Yes, we should add a spec!  I was a little unsure of how to do so, but that's a great suggestion.

I am also trying to keep in mind your conventions for how to name the methods ending in `?` with regards to testing because I think they are very helpful. :-)  I will take a closer look and see what will make more sense!

app/services/message_employee_matcher.rb
ccostino: Will do!

spec/services/message_employee_matcher_spec.rb
jessieay: yes! - for more context on the thinking behind this, this is a good one: https://robots.thoughtbot.com/lets-not

app/views/scheduled_messages/_form.html.erb
ccostino: Thanks!  "Uses" does sound better, will change to that or something similar. :-)

db/migrate/20151021194020_add_time_of_day_to_scheduled_messages.rb
ccostino: Well, yes and no.  Yes in the sense that it is stored as UTC internally on the object, but no in the sense that adding the time to a scheduled message doesn't need to account for the time zone, just a time.  We only care about the time zone in the context of pairing a message being sent to a user.  Does this make sense?

jessieay: ah yes, makes total sense. Maybe in that case we should set the default time for users to EST so people don't get messages w default time at weird times? ps - UGH timezones! :clock1:

jessieay:
Merge when you feel ready - looks :shaved_ice:

app/services/message_employee_matcher.rb
jessieay: ok, let me know if you want any feedback / help on this or testing it

db/migrate/20151021194020_add_time_of_day_to_scheduled_messages.rb
ccostino: Ah good point on the user default!  Yeah time zones (and daylight savings) are a nightmare!

spec/services/message_employee_matcher_spec.rb
ccostino: Ah this is great, thank you!

db/migrate/20151021194020_add_time_of_day_to_scheduled_messages.rb
ccostino: @jessieay:  Despite using both `default` (which should be the correct one to use) and `selected` as options for the form field in the employee form, I cannot get it to recognize using the `Eastern Time (US & Canada)` option as the default.  Do you have any insight into why this might be?

ccostino:
Thank you for all of the help and feedback, @jessieay!  As soon as I can figure out this one last issue with the employee form I will squash and merge. :-)

db/migrate/20151021194020_add_time_of_day_to_scheduled_messages.rb
ccostino: Nevermind, I got this working!

jessieay:
Reminder that we're trying to do this git protocol in this project: https://github.com/18F/dolores-landingham-bot/blob/master/CONTRIBUTING.md#git-protocol (no merge commits) -- if this system isn't working, let's updating the contributing doc!

Issue #97, opened 2015-11-04
Messages were not sent to everyone


ccostino:
After deploying the time zone enhancement for issue #76, we noticed that some messages were not sent today, at least to EST employees.  This needs to be addressed to ensure that all folks receive messages from Dolores when expected.

PR #98, opened 2015-11-04
Connects #97:  Fixes an issue when checking employee local times against the time a scheduled message should be sent.

ccostino:
This changeset addresses an issue that could potentially prevent a message from being sent if the local time of a user does not exactly match (by the hour and minute) the time a scheduled message is set to be sent out.  This adjustment changes the check to look for a match at the current time or any time thereafter.  A separate check is already in place that ensures the message has not already been sent to the employee.

### DIFF
--- a/app/services/message_employee_matcher.rb
+++ b/app/services/message_employee_matcher.rb
@@ -21,10 +21,9 @@ class MessageEmployeeMatcher
   end
 
   def match_time(time_zone)
-    employee_current_hour = Time.current.in_time_zone(time_zone).hour
-    employee_current_minute = Time.current.in_time_zone(time_zone).min
+    employee_current_time = Time.current.in_time_zone(time_zone)
 
-    employee_current_hour == message.time_of_day.hour && employee_current_minute == message.time_of_day.min
+    employee_current_time.utc.strftime("%H%M%S%N") >= message.time_of_day.utc.strftime("%H%M%S%N")
   end
 
   def message_not_already_sent?(employee)
--- a/spec/services/message_employee_matcher_spec.rb
+++ b/spec/services/message_employee_matcher_spec.rb
@@ -3,49 +3,62 @@ require "rails_helper"
 describe MessageEmployeeMatcher do
   describe "#run" do
     it "matches messages to users based on days after start" do
-      Timecop.freeze(Time.current)
+      Timecop.freeze(Time.current) do
+        days_after_start = 3
+        scheduled_message = create(:scheduled_message, days_after_start: days_after_start, time_of_day: Time.current.in_time_zone("Eastern Time (US & Canada)"))
+        employee = create(:employee, started_on: days_after_start.days.ago)
 
-      days_after_start = 3
-      scheduled_message = create(:scheduled_message, days_after_start: days_after_start, time_of_day: Time.current.in_time_zone("Eastern Time (US & Canada)"))
-      employee = create(:employee, started_on: days_after_start.days.ago)
+        matched_employees_and_messages = MessageEmployeeMatcher.new(scheduled_message).run
 
-      matched_employees_and_messages = MessageEmployeeMatcher.new(scheduled_message).run
+        expect(matched_employees_and_messages).to eq(
+            [ employee ]
+        )
+      end
+    end
 
-      expect(matched_employees_and_messages).to eq(
-          [ employee ]
-      )
+    it "matches messages to users if the user's time has advanced past the scheduled message send at time but hasn't been sent yet" do
+      Timecop.freeze(Time.current) do
+        days_after_start = 3
+        half_hour_time_difference_in_seconds = 1800
+        scheduled_message_time = Time.current.in_time_zone("Eastern Time (US & Canada)")
+        scheduled_message_time = scheduled_message_time - half_hour_time_difference_in_seconds
+        scheduled_message = create(:scheduled_message, days_after_start: days_after_start, time_of_day: scheduled_message_time)
+        employee = create(:employee, started_on: days_after_start.days.ago)
 
-      Timecop.return
+        matched_employees_and_messages = MessageEmployeeMatcher.new(scheduled_message).run
+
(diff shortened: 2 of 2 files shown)


app/services/message_employee_matcher.rb
```diff
@@ -24,7 +24,10 @@ def match_time(time_zone)
     employee_current_hour = Time.current.in_time_zone(time_zone).hour
     employee_current_minute = Time.current.in_time_zone(time_zone).min
 
-    employee_current_hour == message.time_of_day.hour && employee_current_minute == message.time_of_day.min
+    # Since time is always ticking, we need to check to see if the employee's
```
jessieay: rather than a comment, can we create a test to reproduce this behavior so we will have a test failure if this behavior ever gets reintroduced?

afeld:
Without a lot of context in the code, the change makes sense to me :+1:

ccostino:
@afeld and @jessieay:  Thank you so much for looking!  In taking @jessieay's recommendation, I uncovered a rather egregious assumption on my part about the individual hour and minute checks.  For example, if a message is scheduled to go out at 1:30 PM, but the current time is 2:10, the check will fail!  The hour is greater in this case for the current time (even when converted to a 24 hour clock), but the minutes according to this check will not be greater than the message time.

I've modified the check some to account for checking the time itself in its entirety (ignoring the date) and added a test case that should capture this scenario explicitly.  I still see a potential issue in the case of messages scheduled to go out around midnight, but messages shouldn't be going out at midnight. :-)

spec/services/message_employee_matcher_spec.rb
```diff
@@ -18,6 +18,26 @@
...
+      scheduled_message = create(:scheduled_message, days_after_start: days_after_start, time_of_day: scheduled_message_time)
+      employee = create(:employee, started_on: days_after_start.days.ago)
+
+      matched_employees_and_messages = MessageEmployeeMatcher.new(scheduled_message).run
+
+      expect(matched_employees_and_messages).to eq(
+          [ employee ]
+      )
+
+      Timecop.return
```
jessieay: rather than doing this, you can wrap this spec in a block eg

```ruby
Timecop.freeze(Time.current) do

...
...

end
```
your call

spec/services/message_employee_matcher_spec.rb
```diff
@@ -18,6 +18,26 @@
       Timecop.return
     end
 
+    it "matches messages to users if the user's time has advanced past the scheduled message send at time but hasn't been sent yet" do
+      Timecop.freeze(Time.current)
+
+      days_after_start = 3
+      scheduled_message_time = Time.current.in_time_zone("Eastern Time (US & Canada)")
+      scheduled_message_time -= 1800
```
jessieay: I generally think it's easier to read something like `scheduled_message_time = scheduled_message_time - 1800` -- thoughts?

spec/services/message_employee_matcher_spec.rb
```diff
@@ -18,6 +18,26 @@
...
+      puts Time.current
```
jessieay: accidentally let these `puts` statements?

spec/services/message_employee_matcher_spec.rb
jessieay: also, does 1800 represent seconds here? Maybe assign as a var to clarify meaning?

spec/services/message_employee_matcher_spec.rb
ccostino: D'oh, yes. :-(

jessieay: I do it allll the time - I think there's a code climate thing you can set up to check for this. we really need to get this project set up on CC....

spec/services/message_employee_matcher_spec.rb
ccostino: This is one case where I tend to veer away from being verbose (as long as it is a simple statement like in this case), but I am fine changing it as several folks have mentioned the same preference about statements like this in previous teams I've been a part of.

app/services/message_employee_matcher.rb
```diff
@@ -21,10 +21,9 @@ def day_count
   end
 
   def match_time(time_zone)
-    employee_current_hour = Time.current.in_time_zone(time_zone).hour
-    employee_current_minute = Time.current.in_time_zone(time_zone).min
+    employee_current_time = Time.current.in_time_zone(time_zone)
 
-    employee_current_hour == message.time_of_day.hour && employee_current_minute == message.time_of_day.min
+    employee_current_time.utc.strftime("%H%M%S%N") >= message.time_of_day.utc.strftime("%H%M%S%N")
```
monfresh: What do you think about converting the dates/times to integers and comparing those instead of comparing strings?

spec/services/message_employee_matcher_spec.rb
ccostino: I did not know this, I didn't see it in the documentation, thanks!

app/services/message_employee_matcher.rb
afeld: You should be able to compare Time objects directly, no?

ccostino: The problem with that in this case is that the date portion of the Time object will not match, nor do we care about it - we just want the time portion of the scheduled message (e.g., the scheduled message is set to go out at 12:00 PM) and compare that to just the current local time of the user.  Comparing the objects directly is going to factor in the date too, and the date in the case of the scheduled message happens to get filled in as 2000-01-01.

spec/services/message_employee_matcher_spec.rb
ccostino: Yes, or as a part of a linter running locally to catch such things!

spec/services/message_employee_matcher_spec.rb
ccostino: Ah good point, addressing that - it's meant to represent a difference of half an hour (so simulating the scheduled message being set to be sent in the past from the current time the test is run).

Issue #100, opened 2015-11-05
Messages were sent out to all employees at the same time this morning


ccostino:
This morning I noticed I had received a message from the bot at 7 AM EST when all of the messages are still scheduled to go out at 12 PM.  Upon looking at the listing of sent scheduled messages, it appears that everyone, regardless of their time zone, received a message at the same time this morning.  It would appear the fix yesterday for issue #97 didn't quite hit the mark, unfortunately.

The tests should have covered this but for some reason didn't.  I am going to try and improve them so that they ensure that messages get sent out exactly as they are expected to at employee's local times and then hopefully uncover the breakdown.

PR #101, opened 2015-11-05
Connects #100:  Fixes issue with messages sent at the wrong time.

ccostino:
The changeset introduced to fix issue #97 had another problem with it in that the time check between the current time of the employee and the time the scheduled message was supposed to go out was still incorrect.  This changeset fixes the time comparison check and bolsters the tests to account for this situation.


jessieay:
:sheep: it!

Issue #38, opened 2015-09-25
Add pagination


jessieay:
Our index views will get huge, eventually. We should add some form of pagination via something simple like Kaminari https://github.com/amatsuda/kaminari

jessieay:
Closing in favor of https://trello.com/c/l37skRw9

PR #102, opened 2015-11-05
Connects #38:  Add pagination to index pages.

ccostino:
This changeset adds support for pagination on all of the index pages.  It also truncates the message body on the sent scheduled messages listing to help with the display of the page.

### DIFF
--- a/config/initializers/kaminari_config.rb
+++ b/config/initializers/kaminari_config.rb
@@ -1,0 +1,10 @@
+Kaminari.configure do |config|
+  config.default_per_page = 10
+  config.max_per_page = nil
+  config.window = 4
+  config.outer_window = 0
+  config.left = 0
+  config.right = 0
+  config.page_method_name = :page
+  config.param_name = :page
+end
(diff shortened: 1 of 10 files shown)


config/initializers/kaminari_config.rb
```diff
@@ -0,0 +1,10 @@
+Kaminari.configure do |config|
+  config.default_per_page = 10
```
jessieay: can we add a quick test where we stub this number to be something like 1 and make sure that a second record doesn't show? Not a huge deal, just would be nice to have a smoke test that pagination is working

ccostino: Can do!

spec/features/view_scheduled_messages_spec.rb
```diff
@@ -18,6 +18,38 @@
     expect(page).to have_content(second_scheduled_message.time_of_day.strftime("%l:%M %p"))
   end
 
+  scenario "sees pagination controls" do
+    allow(Kaminari.config).to receive(:default_per_page).and_return(1)
```
jessieay: awesome

jessieay:
SHIP IT!

Issue #104, opened 2015-11-09
Add support for time and time zone diagnostics


ccostino:
We are still experiencing an issue with scheduled messages not being sent at the correct time and at this point it is unclear as to what the problem is.  Despite the Rails configuration having the timezone set to "UTC" it appears that the server time the app is deployed to could still be used for the time comparison checks, delayed job scheduling, etc.  It's possible we may also need to have the `TZ` environment variable explicitly set to "UTC" (http://stackoverflow.com/questions/4703830/set-time-zone-offset-in-ruby).

At this point we need a bit of a window into what the server and application are reporting for their current time and time zones, so let's add a diagnostics endpoint to see some of this information and hopefully shed some more light on this issue.

For reference:  http://www.elabs.se/blog/36-working-with-time-zones-in-ruby-on-rails

ccostino:
This could also be an issue with using Clockwork, particularly the setting of a scheduled event:  https://github.com/tomykaira/clockwork#tz-1

PR #105, opened 2015-11-09
Connects #104:  Add diagnostic page for time and time zone info.

ccostino:
This changeset adds a diagnostics page that displays application and server information regarding the current time and time zone settings.  It also slightly modifies the time check for matching messages to employees to be a bit more explicit and adds a Clockwork default time zone configuration for UTC.  The hope is that with these changes and additional information being exposed we can finally uncover why the scheduled messages have been sent out at the wrong time.


jessieay:
LGTM!

ccostino:
Thanks, @jessieay!

PR #108, opened 2015-11-13
Rename and refactor match_time

monfresh:
**Why**:
Renaming `match_time` to `time_to_send_message?` better expresses what it does, which is to determine whether or not it's time to send a message. And to answer that question consistently with either true or false (as opposed to true or nil), we use a guard clause at the top to make it clear what the failing scenario is.

If a method returns a boolean, the convention is to make it end with a question mark.

### DIFF
--- a/app/services/message_employee_matcher.rb
+++ b/app/services/message_employee_matcher.rb
@@ -12,21 +12,25 @@ class MessageEmployeeMatcher
   attr_reader :message
 
   def retrieve_matching_employees
-    Employee.where(started_on: day_count.days.ago)
-      .select { |employee| match_time(employee.time_zone) && message_not_already_sent?(employee) }
+    Employee.where(started_on: day_count.days.ago).select do |employee|
+      time_to_send_message?(employee.time_zone) && message_not_already_sent?(employee)
+    end
   end
 
   def day_count
     message.days_after_start
   end
 
-  def match_time(time_zone)
+  def time_to_send_message?(time_zone)
     employee_current_time = Time.current.in_time_zone(time_zone)
 
     if employee_current_time.day == Time.current.day
       employee_current_time_value = employee_current_time.strftime("%H%M").to_i
       message_time_value = message.time_of_day.strftime("%H%M").to_i
+
       employee_current_time_value >= message_time_value
+    else
+      false
     end
   end
 


app/services/message_employee_matcher.rb
```diff
@@ -12,22 +12,24 @@ def run
...
-  def match_time(time_zone)
+  def time_to_send_message?(time_zone)
     employee_current_time = Time.current.in_time_zone(time_zone)
 
-    if employee_current_time.day == Time.current.day
-      employee_current_time_value = employee_current_time.strftime("%H%M").to_i
-      message_time_value = message.time_of_day.strftime("%H%M").to_i
-      employee_current_time_value >= message_time_value
-    end
+    return false if employee_current_time.day != Time.current.day
```
jessieay: I am generally not a fan of  early`return` statements in Ruby (I know they are recommended by some). I think that `if` statements are easier to understand / more readable.

In a situation like this, I would instead do something like:

```ruby
def time_to_send_message?(time_zone)
  employee_current_time = Time.current.in_time_zone(time_zone)

  if employee_current_time.day == Time.current.day
    ...
  end
end
```

jessieay:
One comment about using an early `return`, otherwise looks good to go!

app/services/message_employee_matcher.rb
monfresh: As I mentioned in my commit message, I made this change for reasons beyond readability. With the previous version of the method, when the condition fails, the method will return `nil`. With my change, it will return `false`. I believe a predicate method should return `true` or `false`, not `true` or `nil`. 

Apparently, this is a common debate in the Ruby world with opinions on both sides, but I tend to agree with the viewpoint that a method should represent its intent. See this article for example: http://pragmati.st/2012/03/24/the-elements-of-ruby-style-predicate-methods/

If the intent is to return whether or not it is time to send the message, I think it's clearer to return `false` as opposed to `nil`, and doing that via an early return is better IMO than adding an `else false` at the end.

jessieay: I agree that method should reveal intent. I don't usually worry too much about explicitly returning `false` rather than `nil` from predicate methods, but sounds like a good idea. 

So what about ending with an `else false` here? I think that would be clearer than the return statement and also satisfies the "represent intent" piece.

ccostino:
@monfresh, thanks for taking a look at this and refactoring and thank you @jessieay for reviewing!  I don't know why I didn't think to use a question mark in the name earlier given that the intent is to return a boolean; thinking in JS or Python mode most likely. :-)

As far as what should be returned the method should return `true` or `false` given it is meant to explicitly indicate whether or not it is time to send the message.  `nil` should not be returned here and the fact that the method inadvertently relied on Ruby's truth checks is incorrect in this case.

I'm less concerned about the early return as I've grown accustomed to short circuiting code in cases like this where there is no reason to continue execution if it is not necessary and the composition of the method is easy to digest.

In pulling down this changeset and running it locally things appear to be working fine!  However I uncovered another issue with two failing tests related to  the message_employee_matcher:

* MessageEmployeeMatcher#run matches messages to users based on days after start
* MessageEmployeeMatcher#run ignores messages that have already been sent to users

Both of these test fail on the master branch too (yet for no reason I can see, the last change that modified any related code was the fix @jessieay and I made, which worked!) so I don't think it's a problem with this changeset, however it is making it hard to test the code running locally since it fails during execution as well. :-)

I can help take a look tomorrow, but I think this PR is fine as is!

ccostino:
Just a quick update after trying a couple of things this morning regarding the failing tests:  first, they are definitely unrelated to this change.  Second, I had thought that maybe I needed to have my local machine also running (or report itself as running via the TZ environment variable) in UTC (it's set to EST by default).  I tried this last night but the same issue took place, both of those tests failed and no message would be sent out.

When I tried this again this morning, the tests miraculously all passed and the test messages were being sent fine, but with my machine running in the timezone it normally does and as UTC.

Long story short, in trying to test and verify this change I uncovered another issue when sending out messages late at night.  I am not sure what the actual conditions are, but I'll write a separate issue to cover it.

This PR is good to go! :-)

app/services/message_employee_matcher.rb
monfresh: I disagree that an `else false` is clearer than an early return. With the early return, you can tell right away which condition results in `false`. I think putting the `else false` at the end puts extra cognitive burden on the reader because they have to read through several lines of the `true` scenario, and by the time they get to the `else`, they might have forgotten what the condition at the top was.

If the `if` portion was only one line long, I wouldn't care so much about using `else`, but since the `if` portion is loaded, I think an early return is justified in this case.

jessieay: Perhaps we should have a larger convo about this in the code style guide repo! For now, this project is not using early returns, so I think it would be stylistically consistent to do `else false` -- cool?

monfresh:
No problem, @jessieay. I made the change. I look forward to starting the discussion in the dev guide and seeing what others think.

jessieay:
looks great! merge at your leisure

PR #109, opened 2015-11-16
Add Travis CI config

### LINKED ISSUE #47: Set up CI
Travis or Circle

monfresh:
**Why**:
So that the tests can be run automatically on a continuous integration service.

Note that someone with admin rights to the repo needs to turn on the repo here: https://travis-ci.org/profile/18F

Closes #47.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,16 @@
+before_install:
+- export DISPLAY=:99.0
+- sh -e /etc/init.d/xvfb start
+before_script:
+- cp .sample.env .env
+- bin/rake db:setup --trace
+bundler_args: "--deployment --jobs=3 --retry=3 --without development"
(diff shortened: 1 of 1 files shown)


monfresh:
Since the purpose of this PR is to set up Travis and make sure the build runs, I would say it can be merged. The reason for the failure on Travis is due to the `bundler:audit` reporting a vulnerability with the omniauth-myusa gem. That should probably be addressed in a separate PR. Thoughts?

jessieay:
Your explanation makes sense. One question: why are there two checks present for Travis? 

If that is expected, feel free to merge this PR, @monfresh ! :ship:

monfresh:
Yep, that is expected, albeit confusing: https://github.com/travis-ci/travis-ci/issues/3241

PR #112, opened 2015-11-20
Use full https URL for gems that point to GitHub

monfresh:
**Why**
By using the `github` shortcut, `bundler-audit` reports the omniauth-myusa gem as using an insecure source, and since `bundler-audit` runs as part of the Travis integration, builds are being reported as failed.

To fix this, we need to use the full https URL. See this issue for details:
https://github.com/rubysec/bundler-audit/issues/66

Note that I ran `bundle install` after changing the gem source, which is why Gemfile.lock was changed.


monfresh:
Pinging @jessieay @ccostino for review when you get a chance. Not sure if you saw this come through last week.

jessieay:
Looks good! Merge when ready :ship:

PR #126, opened 2016-01-07
Business Time!

### LINKED ISSUE #19: Holiday Weeks
Hey!

Thought popped into my head this AM: sometimes there's a holiday, in which case I think we'd want to bump messages to the following day. This isn't necessary for v.1 - but just wanted to bring to your attention. :)

https://www.opm.gov/policy-data-oversight/snow-dismissal-procedures/federal-holidays/#url=2015

gemfarmer:
This PR fixes #19, addressing the problem whereby Dolores sent messages on weekends and holidays.

Per @cllns's recommendation, I used the `holidays` and `business_time` gems. They worked great, thanks Sean!

NOTE: one additional question that popped up while I was doing this is whether or not to include regionally-observed holidays on a region-by-region basis. I left it out currently because it was an extra hurdle and wasn't sure how important it was. What do you think @melodykramer ?

Ready for review. @ccostino @jessieay 

It's business time!

### DIFF
--- a/config/business_time.yml
+++ b/config/business_time.yml
@@ -1,0 +1,16 @@
+business_time:
+  beginning_of_workday: 9:00 am
+  end_of_workday: 5:00 pm
+  holidays:
+    - September 28, 2015
+    - September 29, 2015
+    - September 30, 2015
+    - December 24, 2015
+    - May 4, 2016
+    - May 5, 2016
+  work_week:
+    - mon
+    - tue
+    - wed
+    - thu
+    - fri
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -33,6 +33,10 @@ document.
 18f-ers can merge their changes into master after getting approval from another
 contributor.
 
+### Managing time / Updating holidays
+Each message created will send a specified number of **business days** after an employee joins 18F.
+What constitutes a business day is managed by the gem `business_time` and is configured [here](config/initializers/business_time.rb) and [here](config/business_time.yml). To add days that dolores will skip, add that date to the `holidays` field in [this yaml config file](config/business_time.yml).
+
 ### App setup
 
 To get started, run `bin/setup`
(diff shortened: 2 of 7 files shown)


melodykramer:
I'm totally okay with not including regional-specific holidays because I don't think they would affect 18F. For example, I'm pretty sure Boston has a holiday in the spring but any 18F member in Boston would still be working.

gemfarmer:
Thant makes sense to me!

melodykramer:
Leap year good? :)

gemfarmer:
@melodykramer It should be!

cllns:
Awesome :)

config/business_time.yml
```diff
@@ -0,0 +1,16 @@
+business_time:
+  beginning_of_workday: 9:00 am
+  end_of_workday: 5:00 pm
+  holidays:
+    - September 28, 2015
+    - September 29, 2015
+    - September 30, 2015
+    - December 24, 2015
+    - May 4, 2016
+    - May 5, 2016
```
ccostino: Are these just sample values used during testing or examples?  Should we change or remove them?

ccostino:
@gemfarmer:  This is awesome!  @cllns, thank you for the gem suggestions. :-)

I think this is good to go, I just had one note/question around the holidays added to the YAML config.

config/business_time.yml
gemfarmer: Correct! These are not sample or testing values. They are offsite days and the day before christmas (half day). I think it makes sense to exclude offsite days to encourage people to be more present, but open to suggestions. @melodykramer, what do you think?

ccostino: Ah, that is a good point!

gemfarmer:
@ccostino in that case I will squash and merge!

ccostino:
@gemfarmer:  Sounds good!

Issue #128, opened 2016-01-12
Pagination causing problems with display


gemfarmer:
Paginating the list of employees is leading to a few issues:
* not all employees are listed
* others are listed twice

This is annoying because it __looks__ llike a user doesn't exist, but they actually do, so we are tempted to create a new user, but a record with that `slack_username` exists and nothing is changed.

I am proposing that we, in the interest of simple approaches, remove the pagination and add a filter functionality to quickly find users.

@melodykramer @jessieay @ccostino thoughts?

melodykramer:
That works for me!

ccostino:
@gemfarmer:  I'm in favor of this as well; the pagination was added only as a quick stop-gap to prevent the pages from growing too long, but clearly it's not quite working right.  Adding a filter mechanism is definitely a good idea to keep things manageable, with or without pagination!

monfresh:
I'm pretty sure this is due to sorting by a non-unique key (`started_on`). Postgres doesn't guarantee the same sort order when multiple records have the same value for the column that is being used to sort. You can test this theory by sorting by Employee id instead.

jessieay:
ah yes! we should sort by `created_at` instead

ccostino:
D'oh!  Thanks for the clarification, @monfresh!  Perhaps we could reinvestigate this after the filtering is done because it would be helpful as more employees and messages are added, and especially as more messages are sent and recorded.

jessieay:
@gemfarmer is this fixed now?

gemfarmer:
@jessieay The pagination should be gone, but I kept the issue open because based on @monfresh and @ccostino's  comments, it might be a good idea to re-introduce pagination in addition to the filtering. Thoughts?

gemfarmer:
I just reintroduced pagination and pushed it live with #138

PR #129, opened 2016-01-13
Remove pagination, add filtering

gemfarmer:
For #128 
For employees, scheduled messages, sent scheduled messages
* Removes pagination
* adds the ability to filter by parameters that would be useful

current status is available at https://dolores-staging.18f.gov

ready for review @melodykramer @ccostino @jessieay 
I am by no means a skilled rubyist, so you will not insult me if you nit pick syntax or anything.

TODOs: 
* add test for filtering options. 
I'm not too experienced with rspec and my eiti responsibilities are preventing me from doing to much in this regard.
* improve filtering, preferably with js so it is a smoother experience


jessieay:
@gemfarmer ping us when this is ready for review!

app/views/sent_scheduled_messages/index.html.erb
```diff
@@ -1,3 +1,17 @@
+<header class="filter-area">
+  <h3>Filter by</h3>
+  <%= form_tag(sent_scheduled_messages_path, :method => "get", class: "navbar-form", slack_username: "search-form") do %>
```
jessieay: new hash syntax `method: "get"`

app/models/sent_scheduled_message.rb
```diff
@@ -9,4 +9,20 @@ class SentScheduledMessage < ActiveRecord::Base
   validates :sent_on, presence: true
 
   delegate :slack_username, to: :employee
+
+  def self.filter(params)
+  	@employee = Employee.find_by slack_username: params[:slack_username]
+
+    results = self.all.where("lower(message_body) like ?", "%#{params[:message_body].downcase}%")
+
+    if @employee
```
jessieay: this spacing seems weird. also don't need to use `return` here. so:

```ruby
def self.filter(params)
  if @employee
    results.where(employee_id: @employee.id)
  elsif params[:sent_on]
    results.where(sent_on: params[:sent_on])
  end 
end
```

app/models/scheduled_message.rb
```diff
@@ -6,4 +6,16 @@ class ScheduledMessage < ActiveRecord::Base
   validates :title, presence: true
 
   acts_as_taggable
+
+  def self.filter(params)
+
+    results = self.all.where("lower(title) like ?", "%#{params[:title].downcase}%")
+    results = results.where("lower(body) like ?", "%#{params[:body].downcase}%")
+
+    if !params[:tag].blank?
```
jessieay: can do ` if params[:tag]` instead

app/models/scheduled_message.rb
```diff
@@ -6,4 +6,16 @@ class ScheduledMessage < ActiveRecord::Base
   validates :title, presence: true
 
   acts_as_taggable
+
+  def self.filter(params)
+
+    results = self.all.where("lower(title) like ?", "%#{params[:title].downcase}%")
+    results = results.where("lower(body) like ?", "%#{params[:body].downcase}%")
```
jessieay: I find re-assignment like this can get kinda confusing sometimes. what about chaining instead:

```ruby
results = self.all.
  where("lower(title) like ?", "%#{params[:title].downcase}%").
  where("lower(body) like ?", "%#{params[:body].downcase}%")
```

app/models/scheduled_message.rb
```diff
@@ -6,4 +6,16 @@ class ScheduledMessage < ActiveRecord::Base
...
+
+    if !params[:tag].blank?
+      tags = params[:tag].split(",").each { |t| t.strip }
+      results = results.tagged_with(tags, :any => true)
+    end
+    return results
```
jessieay: This method can be re-written along the lines of what I suggested in `sent_scheduled_message.rb`

app/models/employee.rb
```diff
@@ -2,4 +2,16 @@ class Employee < ActiveRecord::Base
...
+
+  def self.filter(params)
+    if params[:slack_username] && params[:started_on].blank?
+      results = self.all.where("slack_username like ?", "%#{params[:slack_username]}%")
+    elsif params[:started_on] && params[:slack_username].blank?
+      results = self.all.where(started_on: params[:started_on])
+    elsif params[:started_on] && params[:slack_username]
+      results = self.all.where(slack_username: params[:slack_username], started_on: params[:started_on])
+    else
+      results = self.all
```
jessieay: no need to assign each of these to the var `results`

app/controllers/sent_scheduled_messages_controller.rb
```diff
@@ -1,5 +1,9 @@
 class SentScheduledMessagesController < ApplicationController
   def index
-    @sent_scheduled_messages = SentScheduledMessage.order(sent_on: :desc).page(params[:page])
+  	if !params[:slack_username].blank? || !params[:message_body].blank? || !params[:sent_on].blank?
```
jessieay: I think this `if` should be bumped two spaces to the left

jessieay: I find that `thing.present?` is easier to read than `!thing.blank?` and they are logically equivalent. what do you think?

app/controllers/sent_scheduled_messages_controller.rb
```diff
@@ -1,5 +1,9 @@
...
+      @sent_scheduled_messages = SentScheduledMessage.filter(params).order(sent_on: :desc)
```
jessieay: what about removing this `if` check and just passing `SentScheduledMessage.filter(params).order(sent_on: :desc)` whether those params are present or not. then, in the `filter` method, we can return `SentScheduledMessage.all` if there are no params present

jessieay: it looks like we are already doing something similar for Employee

app/assets/javascripts/fixedsticky.js
```diff
@@ -0,0 +1,191 @@
+;(function( win, $ ) {
```
jessieay: is the js frm a lib? if it is , let's put it in vendor http://guides.rubyonrails.org/asset_pipeline.html#asset-organization

app/assets/javascripts/application.js
```diff
@@ -1,3 +1,4 @@
 //= require jquery
 //= require jquery_ujs
-//= require_tree .
\ No newline at end of file
+//= require_tree .
+
```
jessieay: extra newline :)

app/assets/javascripts/custom.js
```diff
@@ -0,0 +1,3 @@
+$(function(){
```
jessieay: thoughts on naming this file something more specific? not sure what this lib does, but naming it after what it does would be :+1: :thought_balloon:

app/assets/stylesheets/sass/_colors.scss
```diff
@@ -0,0 +1,5 @@
+$white: #fff;
```
jessieay: niiiiiiice

app/assets/stylesheets/sass/core.scss
```diff
@@ -2,6 +2,10 @@
 @import "neat/neat";
 @import "base/base";
 
+@import "colors";
```
jessieay: is import order important here? if not, might be helpful to alpha order

spec/features/create_employees_spec.rb
```diff
@@ -30,4 +30,19 @@
...
+
+    visit new_employee_path
+    fill_in "Slack username", with: username
+    select "2015", from: "employee_started_on_1i"
+    select "June", from: "employee_started_on_2i"
+    select "1", from: "employee_started_on_3i"
+    select "Eastern Time (US & Canada)", from: "employee_time_zone"
+    click_on "Create Employee"
+
+    expect(page).to have_content("The employee '#{username}' already exists in Dolores")
```
jessieay: nice testing!

app/models/sent_scheduled_message.rb
gemfarmer: I'm trying this out, and I seem to need the explicit return. I arrived at this _slightly_ better syntax:
```
    @employee = Employee.find_by slack_username: params[:slack_username]

    results = self.all.where("lower(message_body) like ?", "%#{params[:message_body].downcase}%")

    if @employee
      employee_id = @employee.id
      results.where(employee_id: employee_id)
    end

    if !params[:sent_on].blank?
      results.where(sent_on: params[:sent_on])
    end

    return results
```

app/models/scheduled_message.rb
gemfarmer: agreed :)

app/models/sent_scheduled_message.rb
jessieay: what happens if you do not use `return`? As I am reading it, I would expect both to return the same

app/controllers/sent_scheduled_messages_controller.rb
gemfarmer: agreed

app/assets/javascripts/custom.js
gemfarmer: I figured since we don't have much js, I would just keep it generic for now, but I'll rename it.

app/assets/stylesheets/sass/_colors.scss
gemfarmer: haha yeah, a little unnecessary

app/assets/stylesheets/sass/core.scss
gemfarmer: the import order is important. I wanted to overwrite any colors definted in neat or bourban, but set them before anything else.

gemfarmer:
@jessieay thanks for the review. I made some changes. now visible on https://dolores-staging.18f.gov

jessieay:
@gemfarmer coolio :wine_glass:  let me know when CI is passing and I'll take another look

app/models/scheduled_message.rb
```diff
@@ -6,4 +6,19 @@ class ScheduledMessage < ActiveRecord::Base
...
+  def self.filter(params)
+    if params[:title].present? || params[:body].present? || params[:tag].present?
+      results = self.all.where("lower(title) like ?", "%#{params[:title].downcase}%")
+        .where("lower(body) like ?", "%#{params[:body].downcase}%")
+
+      if params[:tag].present?
+        tags = params[:tag].split(",").each { |t| t.strip }
+        results = results.tagged_with(tags, :any => true)
+      end
+      return results
```
jessieay: pretty sure this line is unnecessary-- what happens when you remove it?

app/models/employee.rb
```diff
@@ -2,4 +2,12 @@ class Employee < ActiveRecord::Base
   validates :slack_username, presence: true, uniqueness: true, format: { with: /\A[a-z_0-9.]+\z/, message: "Slack usernames can only contain lowercase letters, numbers, underscores, and periods." }
   validates :started_on, presence: true
   validates :time_zone, presence: true
+
+  def self.filter(params)
+    results = self.all.where("slack_username like ?", "%#{params[:slack_username]}%")
+    if params[:started_on].present?
+      results.where(started_on: params[:started_on])
+    end
+    return results
```
jessieay: pretty sure this is unnecessary

app/controllers/scheduled_messages_controller.rb
```diff
@@ -16,7 +16,7 @@ def create
   end
 
   def index
-    @scheduled_messages = ScheduledMessage.order(:days_after_start).page(params[:page])
+      @scheduled_messages = ScheduledMessage.filter(params).order(:days_after_start)
```
jessieay: I think this should be bumped two spaces to the left

app/controllers/employees_controller.rb
```diff
@@ -56,6 +63,10 @@ def employee_params
     params.require(:employee).permit(:slack_username, :started_on, :time_zone)
   end
 
+  def existing_employee(slack_username)
+    !!Employee.where(slack_username: slack_username)
```
jessieay: I am not a huge an of `!!` -- I think `Employee.where(slack_username: slack_username).any?` would be easy to read

ccostino: +1 to @jessieay's remark for readability.

app/models/employee.rb
ccostino: @jessieay:  Do you mean the explicit `return` statement or the whole method?  If just the `return` line, will `results` be what's returned, including if the `if` statement evaluates to true?  I know Ruby will automatically return whatever the result was of the last statement in a method, but sometimes it's just not entirely clear from a readability perspective.

app/models/scheduled_message.rb
gemfarmer: On filter:
screen shot 2016-01-14 at 11 24 59 am

app/controllers/employees_controller.rb
gemfarmer: I removed this method entirely. You were right when you said that that method already existed.

vendor/assets/javascripts/fixedsticky.js
```diff
@@ -0,0 +1,191 @@
+;(function( win, $ ) {
+
+  function featureTest( property, value, noPrefixes ) {
+    // Thanks Modernizr! https://github.com/phistuck/Modernizr/commit/3fb7217f5f8274e2f11fe6cfeda7cfaf9948a1f5
+    var prop = property + ':',
+      el = document.createElement( 'test' ),
+      mStyle = el.style;
+
+    if( !noPrefixes ) {
+      mStyle.cssText = prop + [ '-webkit-', '-moz-', '-ms-', '-o-', '' ].join( value + ';' + prop ) + value + ';';
```
ccostino: Should the `';'` in the join method call be a `':'` instead?  It looks like you're setting a CSS property and value with this statement, right?

app/models/scheduled_message.rb
jessieay: ah interesting this method must return `nil` when `params[:tag].present?` is false. 

well at the very least we can remove the word `return` here :)

vendor/assets/javascripts/fixedsticky.js
gemfarmer: This is a library that I grabbed. It does look a little off. I'm used to the `-webkit-property: value` syntax but it seems to work, so maybe there is something that I'm missing.

jessieay:
this looks good to me. feel free to squash n merge when CI passes / you're happy with how things look, @gemfarmer !

vendor/assets/javascripts/fixedsticky.js
ccostino: Ah okay, I wasn't sure but the line did seem a bit weird to me too.

app/models/scheduled_message.rb
gemfarmer: That is most likely what is happening. Removing `return`

gemfarmer:
@jessieay :sunglasses: . Thanks for spending so much time review this!

vendor/assets/javascripts/fixedsticky.js
ccostino: Oh I see what it's doing, it's adding vendor prefixes to the property it's working with and appending them to the existing string.  That `';'` is actually separating one property/value combination from the next, but I read it to be doing that between the property and value initially.

ccostino:
@gemfarmer:  Thank you for working on this! :-)

PR #135, opened 2016-01-23
[Fixes #122] Fix intermittently failing specs

### LINKED ISSUE #122: Explore intermittently failing specs
This branch failed here: https://travis-ci.org/18F/dolores-landingham-bot/builds/94793132

but then passed on other runs without any code changes. I suspect this is related to timezones.

jessieay:
* These would fail if the specs were run after midnight UTC
* Solution: set Timecop to before midnight UTC.
* The only potential problem with this code implementation is that we
  can't schedule messages after 4pm bc in PT that is past midnight UTC -- is that a problem?
* https://github.com/18F/dolores-landingham-bot/issues/122

### DIFF
--- a/spec/services/message_employee_matcher_spec.rb
+++ b/spec/services/message_employee_matcher_spec.rb
@@ -4,7 +4,7 @@ require "business_time"
 describe MessageEmployeeMatcher do
   describe "#run" do
     it "matches messages to users based on days after start" do
-      Timecop.freeze(Time.current) do
+      Timecop.freeze(Time.parse("11:00:00 UTC")) do
         days_after_start = 3
         scheduled_message = create(:scheduled_message, days_after_start: days_after_start, time_of_day: Time.current.in_time_zone("Eastern Time (US & Canada)"))
         employee_one = create(:employee, started_on: days_after_start.business_days.ago)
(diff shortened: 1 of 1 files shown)


jessieay:
cc @ccostino 

more timezone fun!

cc @melodykramer 

see note on not being able to scheduled messages past 4pm. Is this problematic?

ccostino:
Thanks for take a look at these, @jessieay!  This changeset looks fine and ready to go, but like you said, we might still have an issue with sending messages after 4 PM PST.  Unfortunately it's still not clear why this is happening in the first place.  Perhaps we should keep a note of it somewhere in the tests or code itself, or do you think the GitHub issues are record enough for this?

jessieay:
@ccostino I understand why this is happening. In the service object, we make sure that the current day for the employee is the same as the current dat in UTC:

```
 if employee_current_time.day == Time.current.day
```

we do this because otherwise we run into the problem where message times are being compared on two different days (we added this to fix another bug)

I think we could create a situation where we can send message after 4pm PT, but would have to think about it for a bit...

For now, interested in whether @melodykramer thinks this is a problem. If not, we can add a validation to scheduled messages so that they cannot have a time after 4pm.

melodykramer:
Not a problem! Should I fix the ones after 4?

jessieay:
@melodykramer sure, if you could set the messages that are scheduled for after 4 that would be awesome. ;)

melodykramer:
done!

Issue #53, opened 2015-10-01
Only certain people can create messages


jessieay:
all people w/a GSA email address should be able to auth into Dolores, but we want to whitelist certain people to be able to create messages. Not sure exactly how that's done with our auth setup

ccostino:
Another question I have:  does our auth setup currently have a means of managing roles and/or permissions, or is that something we will have to build into the app itself?

jessieay:
@gemfarmer set up MyUSA, perahps he has some suggestions on this?

melodykramer:
This is super important - I don't think we want a scenario where anyone can go in and change previous messages — particularly because they're designed as a block of material.

jessieay:
agreed - going to come up with the quickest / easiest solution I can think of and go with that for now. Def a top development priority

jessieay:
I think the best plan of attack would be to create users in the database similar to how C2 does it with MyUSA: https://github.com/18F/C2/blob/91231060f7c32e31ea3332368ed0fb86316f01e6/app/controllers/auth_controller.rb#L5

the auth hash contains the email address. Then, we can manually set certain users as admins and only those users will be able to create / edit messages.

Melody, what about creating / updating employees? Should that also be for admins only? and what about testing messages? I'd like to keep the app as open as possible to all users while also providing necessary safeguards.

melodykramer:
I think anyone should be able to update employees - because people might want to duck out of the bot messages, and should be allowed to do so.

Testing messages - everyone as well, I think.

jessieay:
Started work on this here https://github.com/18F/dolores-landingham-bot/compare/jy-1-based-counting?expand=1

melodykramer:
Editing + adding messages = some employees.
Updating employees = all employees.
Testing messages = all employees.

jessieay:
addressed via https://github.com/18F/dolores-landingham-bot/pull/141 -- closing

PR #136, opened 2016-01-23
Find or create user record during auth

jessieay:
* Next step is to give admin users certain abilities
* Part of https://github.com/18F/dolores-landingham-bot/issues/53


jessieay:
Failing specs here will be fixed via https://github.com/18F/dolores-landingham-bot/pull/135

cc @gemfarmer

jessieay:
@ccostino this is ready for review if you have a minute :)

gemfarmer:
@jessieay looks good! 

Though, I'm not totally sure how much this changes things. People already couldn't make changes unless they had a `gsa.gov` email. I guess it is another line of defense? Or could someone hit the api without logging in?

jessieay:
@gemfarmer this PR is making it so that we save users in the DB, but there are no user-facing changes.

My next step is going to be to actually update the permissions based on whether a user is an admin or not so we can give admins specific abilities. Without users in the DB, we couldn't have permission levels because all we had as a token saved in a cookie.

Thanks for the review!

gemfarmer:
@jessieay based on @melodykramer's comment [here](https://github.com/18F/dolores-landingham-bot/issues/53#issuecomment-149877620) I thought we were going to be allowing everyone (with a gsa email) to edit?

jessieay:
@gemfarmer per Melody's comment in the linked issue, we are going to limit *some* actions to admins only, but all people will be able to do other things. :rocket:

gemfarmer:
Ah I see that now. Thanks for the clarification.

PR #143, opened 2016-02-16
Fix cf-ssh adapter issue

gemfarmer:
This is a PR without a connected issue. 

I had consistently been having issues running the `cf-ssh` command to access the database for `dolores-app`. I had been receiving the following error:

```ActiveRecord::AdapterNotSpecified: database configuration does not specify adapter``` 

Diego explained to me that this is because we need to have the services (postgres database instance) that we are using for the app defined in the manifest file and not as a cloud foundry environment variable. 

I made the change and it fixed the issue. @jessieay @ccostino Looking for either of your blessing!

### DIFF
--- a/cf-ssh.yml
+++ b/cf-ssh.yml
@@ -13,5 +13,7 @@ applications:
   memory: 1GB
   name: dolores-app-ssh
   no-route: true
+  services:
+  - dolores-app-db
 command: script/start
 domain: 18f.gov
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: cf-ssh.yml [hunk 1]
```diff
@@ -5,6 +5,7 @@ applications:
   env:
     APPLICATION_HOST: dolores-app.18f.gov
     DEFAULT_URL_HOST: dolores-app.18f.gov
+    DISABLE_SANDBOX_WARNING: true
```
jessieay: we don't need this env var -- it was copied over from C2 but it isn't in use in this app

jessieay:
looks good to merge once unnecessary env var removed! 

Thanks for dealing with the cloud.gov stuff again -- U da best :rainbow:

ccostino:
Awesome, thanks @gemfarmer!

PR #144, opened 2016-02-18
Style overhaul

gemfarmer:
Dolores' appearance has never been a top priority, but it was starting to drive me a little crazy, so a little over the week ago I decided to overhaul the site, with the new 18F Branding color scheme and and US Web Design Standards in mind.

I have pushed it to staging, so you can see what we are working with --> https://dolores-staging.18f.gov

@jessieay @ccostino @melodykramer @andrewmaier @nicoleslaw Take a look and let me know what you think. I have a thick skin, so design critiques are welcome. :)


gemfarmer:
Oops. Travis is failing. :(

ccostino:
@gemfarmer:  Awesome, thank you!  I can't wait to dive in more but definitely a great start from what I see!

nicoleslaw:
Nice! Thank you so much for doing this.

I think the page that is most important to me as an admin is the edit message page. I can't see that in the staging site (maybe because of the build)—but happy to give you feedback on a screenshot, etc.

gemfarmer:
@ccostino definitely still a lot to improve!
@nicoleslaw I don't think that page will have changed too much – I barely touched the form styling. It should be pretty consistent with the rest of the redesign:


screen shot 2016-02-19 at 10 06 36 am

jessieay:
hah WOAH! I somehow commented on the wrong PR before. Did you notice?

https://github.com/18F/C2/pull/955#issuecomment-186309635

original comment: 

###

Hey @gemfarmer - wow looks so much better!

Comment on implementation - have you considered using the USWDS gem? That would remove the need to have a lot of this boilerplate code in version control (and it has a dependency on bourbon and neat so we wouldn't need those, either).

The released version of the gem has some isssues with Rails / asset pipeline, but I have a fork that works: https://github.com/18F/e-manifest/blob/master/Gemfile#L39

to see how @jeremiak moved e-manifest to USWDS gem, see: 18F/e-manifest#55

If you think it best to do that another time, that's cool too.

###


response to your response:

oh I see! I assumed we were using USWDS because we talked about that at some point in the past. I am fine with *not* using them. I would assume that the 18F style guidelines and the USWDS work nicely together, but that is just an assumption.

gemfarmer:
@jessieay It is a completely reasonable assumption that I would add them after we talked about it. I didn't add USWDS mostly because I thought it was overkill, but if there is a gem, then perhaps I could make a second stab at this with USWDS css and js??

gemfarmer:
It shouldn't be a terribly difficult/long substitution process.

gemfarmer:
@jessieay Would you prefer to merge this now and add USWDS later (with the 18F branding color scheme and logo) or wait until the USWDS gem is added and mark up updated? 

I am leaning towards the former just because there is a lot of restructuring in this branch and it would be nice to merge it and work from that new baseline.

gemfarmer:
Not building properly on this branch. Moving to #145

Issue #142, opened 2016-02-16
Ability to delete a message


nicoleslaw:
As the onboarding lead, I need to be able to delete messages that are no longer important or relevant (or, in today's case, added by mistake).

jessieay:
To clarify for whoever works on this (probably future me) - this should be possible for users who are admins only.

jbcden:
Hello! I thought I'd try my hand at this issue but I am having trouble getting setup. I have been following the instructions in the contributing documentation; however, I am getting a message stating that: `The requested scope is invalid, unknown, or malformed.` I am definitely being redirected to `/auth/myusa` by the `authenticate!` action but I'm not sure why I am failing to be authenticated. I created an application in MyUsa and added the generated keys to the `.env` file. Any thoughts? I would really love to be able to contribute!
PS I also changed then `is_permitted` method in the `AuthController` to accept non-.gov emails and my `.foreman` file does have the port set to 5000. Thanks in advance!

SukhGrewal:
Assume this is for Slack
Would you consider our product VEOCI that already has these and other features that teams need? WE developed Veoci with ten years of experience at GE on what teams need to get their work done. The ability to keep a messaging space clean was one of the first things we did - in real work, it is needed. For example, if someone types in a wrong IP address, you want that corrected and not mislead people for the next year.
VEOCI is a different paradigm from Slack.  But you can get your work done including delete and edit messages if you have the authorization. It is a messaging backbone for work.
For a small onboarding application, we could set you up in minutes.

jessieay:
Hi @SukhGrewal - our team already uses Slack so we are not looking to develop on onboarding application on a different platform.

jbcden:
@jessieay, @nicoleslaw I have opened https://github.com/18F/dolores-landingham-bot/pull/147 to address this issue. Please let me know if there are any problems.

jessieay:
@nicoleslaw this is live on prod!

nicoleslaw:
:+1:

jessieay:
big thank you to @jbcden for stepping in to volunteer on this one. Woot woot! :rainbow:

gemfarmer:
Thanks @jbcden !

PR #147, opened 2016-02-27
Add ability to delete messages

jbcden:
Completes: https://github.com/18F/dolores-landingham-bot/issues/142

### DIFF
--- a/app/controllers/scheduled_messages_controller.rb
+++ b/app/controllers/scheduled_messages_controller.rb
@@ -38,10 +38,10 @@ class ScheduledMessagesController < ApplicationController
   end
 
   def destroy
-    @scheduled_message = ScheduledMessage.find(params[:id])
-    @scheduled_message.destroy
+    scheduled_message = ScheduledMessage.find(params[:id])
+    scheduled_message.destroy
 
-    flash[:notice] = "You deleted #{@scheduled_message.title}"
+    flash[:notice] = "You deleted #{scheduled_message.title}"
     redirect_to scheduled_messages_path
   end
 
--- a/app/controllers/scheduled_messages_controller.rb
+++ b/app/controllers/scheduled_messages_controller.rb
@@ -37,6 +37,14 @@ class ScheduledMessagesController < ApplicationController
     end
   end
 
+  def destroy
+    @scheduled_message = ScheduledMessage.find(params[:id])
+    @scheduled_message.destroy
+
+    flash[:notice] = "You deleted #{@scheduled_message.title}"
+    redirect_to scheduled_messages_path
+  end
+
   private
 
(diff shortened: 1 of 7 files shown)


app/views/scheduled_messages/index.html.erb
```diff
@@ -46,6 +46,12 @@
             <span>Test</span>
           <% end %>
         </td>
+        <td>
+          <a class="button button-delete" data-confirm="Are you sure you want to delete this message?" rel="nofollow" data-method="delete" href="/scheduled_messages/<%= message.id %>">
```
jessieay: nice call on the js confirm!

let's use a rail `link_to` helper rather than an `a` tag:

http://api.rubyonrails.org/classes/ActionView/Helpers/UrlHelper.html#method-i-link_to

and let's use the named rails route instead of creating the URL manually: 

http://api.rubyonrails.org/classes/ActionDispatch/Routing.html#module-ActionDispatch::Routing-label-Named+routes

spec/features/delete_scheduled_messages_spec.rb
```diff
@@ -0,0 +1,16 @@
+require "rails_helper"
+
+feature "Delete scheduled messages" do
+  scenario "from list of scheduled messages", :js do
```
jessieay: I think we only want admins (user with `admin` attribute set to true) to be able to delete messages. Let's write a test for that and implement it!

spec/features/delete_employees_spec.rb
```diff
@@ -6,13 +6,10 @@
 
     login_with_oauth
     visit employees_path
+    click_accept_on_javascript_popup do
     page.find(".button-delete").click
```
jessieay: nice! indent the line within the block?

app/controllers/scheduled_messages_controller.rb
```diff
@@ -37,6 +37,14 @@ def update
     end
   end
 
+  def destroy
+    @scheduled_message = ScheduledMessage.find(params[:id])
```
jessieay: this does not need to be an instance var since we don't reference it in the view. Can just be `scheduled_message` :)

jessieay:
Thank you for this awesome PR! I have a handful of comments. Let me know if any of them need clarification. After those are addressed, I will squash n' merge!

jbcden:
@jessieay Thanks for the comments! They are addressed in the fixups. I also added one commit to the end that changes the employees index page to use the `link_to` helper as well for its deletion link. Let me know if there are any other tweaks to be made.

jessieay:
squashed and merged at befad14929aac63be0f607ab52981b823ca11140 -- thank you very much for your contribution!

Issue #146, opened 2016-02-26
Setup troubles


jbcden:
Hello! I thought I'd try my hand at contributing to [this](https://github.com/18F/dolores-landingham-bot/issues/142) but I am having trouble getting setup. I have been following the instructions in the contributing documentation; however, I am getting a message stating that: `The requested scope is invalid, unknown, or malformed.` I am definitely being redirected to `/auth/myusa` by the `authenticate!` action but I'm not sure why I am failing to be authenticated. I created an application in MyUsa and added the generated keys to the `.env` file. Any thoughts? I would really love to be able to contribute!
PS I also changed then `is_permitted` method in the `AuthController` to accept non-.gov emails and my `.foreman` file does have the port set to 5000. Thanks in advance!

jbcden:
I am also seeing
```
Error parsing header X-XSS-Protection: 1; mode=block, 1; mode=block: expected semicolon at character position 14. The default protections will be applied.
```
when I look in the Chrome console.

jessieay:
Hi @jbcden ! Thank you for chiming in and helping out with the bot. Getting you up and running will help ensure the docs are clear / comprehensive for future developers who want to get going on Dolores, so it's a win-win from our perspective. Apologies if it is frustrating for you. 

One question about your MyUSA app: did you check the "email" box when setting up your myusa app? That is the scope we are asking for and it sounds like the error you're seeing is related to myusa. 

Also, if you are running your app on port 5000, did you add the correct URLs to our myusa app setup?

This is what I have in myusa:

![screen shot 2016-02-26 at 10 24 51 am](https://cloud.githubusercontent.com/assets/601515/13361386/3e4b82c6-dc73-11e5-9332-b214aafd3a3d.png)

If you confirm that all of these settings are working, we can work together to debug more.

jbcden:
@jessieay aha! I had not checked the email box and I also had the first url as just: `http://localhost:5000`. Thanks so much for the quick reply. I would be happy to submit a PR to update the documentation. I think the main points of confusion were the urls and checking the email scope.

gemfarmer:
@jbcden based on your recent PRs, I imagine this is no longer an issue? Also, as Jessie already said, thanks for contributing!

jbcden:
@gemfarmer I am all set now. I will open a PR later this week to update the documentation.

gemfarmer:
@jbcden gotcha. I'll leave this issue open then.

jessieay:
[...]

PR #152, opened 2016-03-04
Update contributions guide.

jbcden:
Addresses issues brought up in https://github.com/18F/dolores-landingham-bot/issues/146.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -85,7 +85,7 @@ development with the following:
 
   For Url:
 
-  `http://localhost:5000/auth/myusa/callback`
+  `http://localhost:5000/`
 
   For Redirect uri:
 
@@ -94,8 +94,9 @@ development with the following:
 2. Under `Select the API Scopes that your Application will use`: select `Email
    Address`.
 
-3. Generate a set of keys, `MYUSA_KEY` and `MYUSA_SECRET`, and reference them in
-`.env`
+3. Generate a set of keys by clicking `New Api Key` next to your application in MyUSA.
+   They will be called `Consumer Public Key` and `Consumer Secret Key` on MyUSA but will
+   map to `MYUSA_KEY` and `MYUSA_SECRET`, in your local `.env` file.
 
 4. Edit the method `is_gsa` and its invocation in
    `/app/controllers/auth_controller` to accomodate non-`gsa.gov` email
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -85,25 +85,28 @@ development with the following:
 
   For Url:
 
-  `http://localhost:5000/`
+  `http://localhost:5000/auth/myusa/callback`
 
   For Redirect uri:
 
   `http://localhost:5000/auth/myusa/callback`
 
-2. Generate a set of keys, `MYUSA_KEY` and `MYUSA_SECRET`, and reference them in
+2. Under `Select the API Scopes that your Application will use`: select `Email
+   Address`.
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -85,25 +85,28 @@ development with the following:
 
   For Url:
 
-  `http://localhost:5000/`
+  `http://localhost:5000/auth/myusa/callback`
```
jessieay: contrary to what I have in my own app setup, this URL does not actually need have this value. Previous value was fine

Sorry for not clarifying that earlier (that being said, having this value doesn't break system, but it isn't necessary)

jbcden: Good to know. The issue I was having must have been due to not having the correct scope set before then. I changed both at the same time without checking to see if only changing one fixed the problem. I will update this accordingly.

CONTRIBUTING.md
```diff
@@ -85,25 +85,28 @@ development with the following:
...
 
   For Redirect uri:
 
   `http://localhost:5000/auth/myusa/callback`
 
-2. Generate a set of keys, `MYUSA_KEY` and `MYUSA_SECRET`, and reference them in
+2. Under `Select the API Scopes that your Application will use`: select `Email
+   Address`.
+
+3. Generate a set of keys, `MYUSA_KEY` and `MYUSA_SECRET`, and reference them in
```
jessieay: maybe we should:

1) tell peoplee they need to click "New API Key" next to the application they just created in myusa in order to create these keys
2) clarify what these are called in myusa? (Consumer Public Key, Consumer Secret Key)

this is perhaps overkill in terms of instructions but would be nice to have super explicit instructions. what do you think?

jessieay:
thanks for submitting this! docs are so important :rainbow:

CONTRIBUTING.md
jbcden: I totally agree. I believe that doing everything possible to lower the barrier to entry for new contributors is the best thing for an open source project.

jbcden: Maybe we should also consider adding a some more detail/links to resources about creating slack bots? Thoughts?

jessieay: I like that idea!

jbcden: Is there some kind of guest/limited access account to the 18f slack? I could see it being annoying for people to have to create a new slack organization to test things. Would it be feasible to enable contributors to have access to a special contributors room under the 18f slack with the ability to add bots? I haven't really messed around with managing a slack organization but this could be just another way to lower the barrier of entry for those who might not have used slack before and want to get into this project quickly. Do you see this as at all feasible?

gemfarmer: Hmm good question. @afeld @wslack @NoahKunin do any of you know if it is possible to have a public slack environment for guests who are interested in using and contributing to our bots?

afeld: Yep! https://handbook.18f.gov/slack/#heading-15

jessieay: I know that for myself when I first asked for an API key for the bot I had to have an 18F Slack admin create one for me. 

My guess is that people working on this will, unfortunately, need to create their own Slack org

jessieay:
it seems like 

> I could see it being annoying for people to have to create a new slack organization to test things

(https://github.com/18F/dolores-landingham-bot/pull/152#discussion_r55098145)

is an open question / issue but can be explored separately. 

Thanks for another stellar contribution and follow up @jbcden

PR #153, opened 2016-03-05
Use OmniAuth MyUSA gem

jessieay:
* https://rubygems.org/gems/omniauth-myusa
* Wasn't on rubygems until today :)


jessieay:
ping ping @ccostino @gemfarmer

ccostino:
@jessieay:  This is just updated the gem reference to grab whatever is in RubyGems instead of directly on GitHub?  If so, looks good to me!

jessieay:
@gemfarmer yup! gem was previously not on rubygems so it was referenced via github. now it is! so we can track versioning etc more easily :tada: 

thanks for the :eyes:

Issue #151, opened 2016-03-03
Sort table by date and time


nicoleslaw:
As the onboarding lead, I need to be able to see the messages at a glance. Right now, they are sorted by date, but the time is ~random. Example below:

![screen shot 2016-03-03 at 10 16 28 am](https://cloud.githubusercontent.com/assets/94900/13498727/271363fc-e129-11e5-93dd-7d72abf0dce9.png)

jessieay:
Hey @nicoleslaw - agreed, messages should be default be sorted in ascending `day_count` and `time` order. 

@gemfarmer weren't you working on something related to sorting records at one point?

gemfarmer:
@jessieay I was, but I wasn't getting this specific. I can't get to this next week, but I can put it on my todo list and get to it asap, @nicoleslaw

jessieay:
cool. I think the first step would be to update default sorting to be ascending day count AND time. second step could be more complicated sorting stuff.

gemfarmer:
Agreed. I'll approach the default first and merge it as its own PR, then if all goes well we can add some sort of sort UI.

jbcden:
@gemfarmer I could implement the default sort if that would be helpful.

gemfarmer:
@jbcden if you have the time, that would be great! Thanks for pitching in!

jessieay:
this is now live on production -- thank you @jbcden for your continued contributions!

PR #154, opened 2016-03-07
Sort by date and time

jbcden:
Completes https://github.com/18F/dolores-landingham-bot/issues/151

### DIFF
--- a/app/controllers/scheduled_messages_controller.rb
+++ b/app/controllers/scheduled_messages_controller.rb
@@ -18,7 +18,8 @@ class ScheduledMessagesController < ApplicationController
   end
 
   def index
-    @scheduled_messages = ScheduledMessage.filter(params).order(:days_after_start).page(params[:page])
+    @scheduled_messages = ScheduledMessage.filter(params)
+                                          .page(params[:page])
   end
 
   def edit
--- a/app/controllers/scheduled_messages_controller.rb
+++ b/app/controllers/scheduled_messages_controller.rb
@@ -19,6 +19,7 @@ class ScheduledMessagesController < ApplicationController
 
   def index
     @scheduled_messages = ScheduledMessage.filter(params)
+                                          .date_time_ordering
                                           .page(params[:page])
   end
 
--- a/spec/features/view_scheduled_messages_spec.rb
+++ b/spec/features/view_scheduled_messages_spec.rb
@@ -11,11 +11,11 @@ feature "View scheduled messages" do
     expect(page).to have_content(first_scheduled_message.title)
     expect(page).to have_content(first_scheduled_message.body)
     expect(page).to have_content(first_scheduled_message.days_after_start)
-    expect(page).to have_content(get_formatted_time(first_scheduled_message))
+    expect(page).to have_content(formatted_time(first_scheduled_message))
     expect(page).to have_content(second_scheduled_message.title)
     expect(page).to have_content(second_scheduled_message.body)
     expect(page).to have_content(second_scheduled_message.days_after_start)
-    expect(page).to have_content(get_formatted_time(second_scheduled_message))
+    expect(page).to have_content(formatted_time(second_scheduled_message))
   end
 
   scenario "sees pagination controls" do
@@ -30,7 +30,7 @@ feature "View scheduled messages" do
     expect(page).to have_content(first_scheduled_message.title)
     expect(page).to have_content(first_scheduled_message.body)
     expect(page).to have_content(first_scheduled_message.days_after_start)
-    expect(page).to have_content(get_formatted_time(first_scheduled_message))
+    expect(page).to have_content(formatted_time(first_scheduled_message))
 
     expect(page).not_to have_content(second_scheduled_message.title)
 
@@ -44,7 +44,7 @@ feature "View scheduled messages" do
     expect(page).to have_content(second_scheduled_message.title)
     expect(page).to have_content(second_scheduled_message.body)
     expect(page).to have_content(second_scheduled_message.days_after_start)
-    expect(page).to have_content(get_formatted_time(second_scheduled_message))
+    expect(page).to have_content(formatted_time(second_scheduled_message))
 
     expect(page).to have_content("Prev")
     expect(page).to have_content("First")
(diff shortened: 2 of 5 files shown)


app/controllers/scheduled_messages_controller.rb
```diff
@@ -18,7 +18,8 @@ def create
   end
 
   def index
-    @scheduled_messages = ScheduledMessage.filter(params).order(:days_after_start).page(params[:page])
+    @scheduled_messages = ScheduledMessage.filter(params)
```
jessieay: should we be using `date_time_ordering` here? 

also, let's write a feature test for this so we have a broken test if sorting breaks!

consider using an approach like this: http://stackoverflow.com/a/10847756/1019369

spec/models/scheduled_message_spec.rb
```diff
@@ -12,4 +12,35 @@
...
+    let!(:message3) do
+      ScheduledMessage.create!(title: 'test3',
+                               body: 'this is the third test',
+                               days_after_start: 1,
+                               time_of_day: '2000-01-01 11:00:00 UTC',
+                               tag_list: tags)
+    end
+
+    it 'should display the messages in chronological order' do
+      expect(ScheduledMessage.date_time_ordering.first).to eq(message3)
```
jessieay: how about using `match_array` here rather than just looking at the first element? that way we know the ordering is working for all items.

also, we are mostly not using `let` statements in this test suite. some explanation of why: https://robots.thoughtbot.com/lets-not

so can we create the scheduled messages in the test itself rather than in `let` statements? Using the `scheduled_message` factory would be great, too.

app/controllers/scheduled_messages_controller.rb
jbcden: Oops, yes, not sure how that happened. Fixed in 5aa8be4c2f0bb5e3ee419081d7bd5b92eb4d3f00

jbcden:
Spec related comments addressed in b3efe4a87e0a764425eadf6454fd1fe10cae041a

spec/models/scheduled_message_spec.rb
```diff
@@ -12,4 +12,13 @@
      it { should validate_presence_of(:time_of_day) }
      it { should validate_presence_of(:title) }
   end
+
+  describe '.date_time_ordering' do
+    it 'should display the messages in chronological order' do
+      m1 = create(:scheduled_message, time_of_day: '2000-01-01 16:00:00 UTC')
+      m2 = create(:scheduled_message, time_of_day: '2000-01-01 14:00:00 UTC')
+      m3 = create(:scheduled_message, time_of_day: '2000-01-01 11:00:00 UTC')
+      expect(ScheduledMessage.date_time_ordering).to match_array([m1, m2, m3])
```
jessieay: nice!

jessieay: perhaps we should add another scheduled message for a different day here to make sure date AND time ordering are working?

spec/features/view_scheduled_messages_spec.rb
```diff
@@ -44,24 +44,44 @@
...
+
+    expect(page).to have_selector("table > tr:nth-child(2) > td:nth-child(2)",
+                                  text: get_formatted_time(first_scheduled_message).strip)
+    expect(page).to have_selector("table > tr:nth-child(3) > td:nth-child(2)",
+                                  text: get_formatted_time(second_scheduled_message).strip)
+  end
+
   private
 
+  def get_formatted_time(scheduled_message)
```
jessieay: what about naming this `formatted_time` instead of `get_formatted_time` ? seems like it is as clear and shorter!

spec/features/view_scheduled_messages_spec.rb
```diff
@@ -44,24 +44,44 @@
...
   end
 
+  scenario "see message in the correct order" do
+    login_with_oauth
+    visit root_path
+    create_scheduled_messages
+
+    visit scheduled_messages_path
+
+    expect(page).to have_selector("table > tr:nth-child(2) > td:nth-child(2)",
```
jessieay: sometimes when I have tricky expectations like this I like to wrap them in well-named methods that reveal their intentions, so for example you could do:

```ruby
expect(page).to have_message_in_order(message: first_scheduled_message, order: 1)
expect(page).to have_message_in_order(message: second_scheduled_message, order: 2)
```

```ruby
def have_message_in_order(message:, order: )
  order = order + 1
  have_selector("table > tr:nth-child(#{order}) > td:nth-child(2)",
    text: get_formatted_time(message).strip)
end
```
what do you think?

jessieay: (note: method would be a private method like the others you've defined here)

jbcden:
@jessieay I addressed your comments in 09b3407 and 86b1eb7. Thanks for the awesome feedback!

jbcden:
I also noticed that `spec/services/message_employee_matcher_spec.rb:37` seems to be failing. It looks like this might be due to the recent time change. I am not exactly sure what the best practice around this would be but I added a possible fix in 07f1c44.

spec/models/scheduled_message_spec.rb
```diff
@@ -12,4 +12,15 @@
...
   end
+
+  describe '.date_time_ordering' do
+    it 'should display the messages in chronological order' do
+      m1 = create(:scheduled_message, days_after_start: 1, time_of_day: '2000-01-01 11:00:00 UTC')
+      m2 = create(:scheduled_message, time_of_day: '2000-01-01 16:00:00 UTC')
+      m3 = create(:scheduled_message, time_of_day: '2000-01-01 14:00:00 UTC')
+      m4 = create(:scheduled_message, time_of_day: '2000-01-01 11:00:00 UTC')
+      m5 = create(:scheduled_message, days_after_start: 4, time_of_day: '2000-01-01 11:00:00 UTC')
+      expect(ScheduledMessage.date_time_ordering).to match_array([m1, m2, m3, m4, m5])
```
jessieay: I am going to merge this as-is but wanted to leave a comment for future consideration:

in specs it is helpful to create messages in *not* the order you want them returned, to make sure that the ordering is really working and not just in the default order (which is `created_at`).

jessieay:
@jbcden updates look amazing! Squashed and merged at 5d52cf5179f4620f01332749c5eb7ec73268551f

Thank you for another great contribution!

Issue #81, opened 2015-10-19
no @ is required for Slack names


jessieay:
make note of this in docs and/or form for adding employees

melodykramer:
One other thing: no spaces can be after the username, either. :)

jessieay:
to implement this, we should add a form validation that does not allow an `@` or a space after the username when filling out the employee#new form (or creating data in general)

jbcden:
@jessieay perhaps I am misunderstanding but doesn't https://github.com/18F/dolores-landingham-bot/blob/master/app/models/employee.rb#L6 do what you are describing? Or do you mean more on the front-end side of things?

jessieay:
yeah I think more of a frontend prompt. I believe right now you get an activerecord validation if the format is wrong but it would be nice to let a person creating an `employee` that the `slacker_username` field should contain the username without an `@`. clients-side validations would also be cool

jbcden:
If no one else gets to this, I would be happy to work on it.

jbcden:
While working on this I noticed that HTML5 validations were disabled in the `simple_form` configurations; is there a reason for this? I was planning on using those but if you have disabled them for a reason I can just use some javascript. Thoughts @jessieay?

jessieay:
@jbcden I am not sure why they were disabled...fine with adding them back in unless others object

PR #155, opened 2016-03-08
New employee form validations

jbcden:
Completes https://github.com/18F/dolores-landingham-bot/issues/81

### DIFF
--- a/app/controllers/employees_controller.rb
+++ b/app/controllers/employees_controller.rb
@@ -7,7 +7,9 @@ class EmployeesController < ApplicationController
   def create
     @employee = Employee.new(employee_params)
 
-    if unknown_employee(@employee.slack_username)
+    if !@employee.valid?
+      render action: :new
+    elsif unknown_employee(@employee.slack_username)
       flash.now[:error] = "There is not a slack user with the username \"#{@employee.slack_username}\" in your organization."
       render action: :new
     elsif @employee.save
--- a/app/controllers/employees_controller.rb
+++ b/app/controllers/employees_controller.rb
@@ -7,19 +7,11 @@ class EmployeesController < ApplicationController
   def create
     @employee = Employee.new(employee_params)
 
-    if !@employee.valid?
-      render action: :new
-    elsif unknown_employee(@employee.slack_username)
-      flash.now[:error] = "There is not a slack user with the username \"#{@employee.slack_username}\" in your organization."
-      render action: :new
-    elsif @employee.save
+    if @employee.save
       flash[:notice] = "Thanks for adding #{@employee.slack_username}"
       redirect_to root_path
-    elsif !unknown_employee(@employee.slack_username)
-      flash.now[:error] = "There is already a slack user with the username \"#{@employee.slack_username}\" in your organization."
-      render action: :new
-    elsif
-      flash.now[:error] = "Could not create employee"
+    else
+      flash.now[:error] = @employee.errors.full_messages.to_sentence
       render action: :new
     end
   end
--- a/app/views/employees/_form.html.erb
+++ b/app/views/employees/_form.html.erb
@@ -1,5 +1,11 @@
 <%= simple_form_for @employee do |form| %>
-  <%= form.input :slack_username %>
+  <%= form.input :slack_username,
+                 required: true,
+                 pattern: '^[^@]\w+[^ ]$',
+                 input_html: {
+                   title: I18n.t('employees.errors.slack_username')
+                 }
+  %>
   <%= form.input :started_on, input_html: { class: "date-time-select" }, order: [:month, :day, :year] %>
   <%= form.input :time_zone, priority: /US/ %>
   <%= form.button :submit %>
--- a/app/models/employee.rb
+++ b/app/models/employee.rb
@@ -3,7 +3,17 @@ class Employee < ActiveRecord::Base
 
   has_many :sent_scheduled_messages, dependent: :destroy
 
-  validates :slack_username, presence: true, uniqueness: true, format: { with: /\A[a-z_0-9.]+\z/, message: "Slack usernames can only contain lowercase letters, numbers, underscores, and periods." }
+  validates :slack_username,
+    presence: true,
+    uniqueness: true,
+    inclusion: {
+      in: EmployeeFinder.new
+            .users_list
+            .map { |h| h['name'] },
+      message: "There is not a slack user with the username \"%{value}\" in"\
+        " your organization."
+    },
+    format: { with: /\A[a-z_0-9.]+\z/, message: "Slack usernames can only contain lowercase letters, numbers, underscores, and periods." }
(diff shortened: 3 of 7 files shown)


app/views/employees/_form.html.erb
```diff
@@ -1,5 +1,11 @@
 <%= simple_form_for @employee do |form| %>
-  <%= form.input :slack_username %>
+  <%= form.input :slack_username,
+                 required: true,
+                 pattern: '^[^@]\w+[^ ]$',
+                 input_html: {
+                   title: I18n.t('employees.errors.slack_username')
```
jessieay: can we write a feature spec that confirms that this shows up?

Also, did you think about using the default `simple_form` locales instead? https://github.com/plataformatec/simple_form#i18n

If you put the value in the right yml location, I believe you don't need to explicitly include it in the form.

jbcden: Thanks for the tip about `simple_form` I haven't really used it before; however, it seems that there are only a few different elements that you can do that with but `title` isn't one of them. As I said, I am fairly unfamiliar with `simple_form` so if anyone knows otherwise please let me know.

jbcden: Also, I am not exactly sure of how to assert on the HTML5 validation, any suggestions?

jessieay: it seems like if you added this message as a custom message to the validation in the `Employee` class, it should show up if the validation fails

see: http://stackoverflow.com/a/11275441/1019369

I am not sure how to assert validations in a test -- perhaps difficult in which case don't worry about it :cherries:

jbcden: Looking at that thread, it seems that that is to ensure HTML4 compatibility. There is actually already a message specified in the model. I think the reason we aren't seeing it is the way the controller handles the `create` action. Instead of checking the validations, we are instead going into a large conditional where we are triggering the first item and immediately failing as can be seen [here](https://github.com/18F/dolores-landingham-bot/blob/master/app/controllers/employees_controller.rb#L10) -- this instead presents a message about no such username existing in your slack organization. We could remove all of the HTML5 validations by reordering the conditional a bit. I will submit a commit with those changes and you can let me know which you prefer.

jbcden: I implemented my idea in 86738f9. I was also wondering if we need the validation on [line 16](https://github.com/18F/dolores-landingham-bot/blob/master/app/controllers/employees_controller.rb#L16). Perhaps there is an edge case I am not grasping but it seems unnecessary to reject users that are already in your organization if they fail to save. I think the check I added in my last commit should handle any dangling edge cases. Thoughts?

jessieay: Yes now that I revisit the logic, I think you are onto something.

Rather than checking for whether a Slack username exists in our org in the controller via `unknown_employee(@employee.slack_username)`, we should add a validation in the `Employee` model that ensures the Slack username is included in the list of all slack usernames for the current org (http://guides.rubyonrails.org/active_record_validations.html#inclusion)

We could also add a uniqueness validation to `slack_username` so we wouldn't have to worry about `flash.now[:error] = "There is already a slack user with the username \"#{@employee.slack_username}\" in your organization."`

The way it is written currently *does* work but if we just did one check that would be sooo much better. something like:

```ruby
if @employee.save
  flash[:notice] = "Thanks for adding #{@employee.slack_username}"
  redirect_to root_path
else
   flash.now[:error] = @employees.errors.messages.to_sentence
      render action: :new
end
```

what do you think?

jbcden: Sorry it has taken me so long to get back to this! I like that a lot. I uncertain why the check on line 18 is needed. The only way we could get to that check is after saying: The record is valid, this username exists in your slack organization, the record failed to save, oh look, this username exists in your slack organization; which is exactly what we want to happen in the second branch of the conditional. I don't understand why what is considered valid in one context is suddenly an error further down in the conditional tree. 

Perhaps I am missing something, but it seems like the inclusion validation combined with the uniqueness validation on `Employee#slack_username` should be sufficient here. Am I missing a case?

jessieay: I think your analysis is correct. I think this is the way it is because we overlooked the fact that this should be an AR validation rather than controller logic. What we have now (https://github.com/18F/dolores-landingham-bot/blob/5c5b06043a5b08d139fc5d08560c2d6a0f26f894/app/controllers/employees_controller.rb#L7) is def non-ideal and I think this branch would be a great opp to improve upon it!

No worries on timing, I know how it can be finding time to work on side stuff! :)

jbcden: I just pushed 30d1f19 and 29c03ab, but I am getting a webmock error in `spec/models/employee_spec.rb`. I'm not sure why webmock isn't catching it. My only guess is it has something to do with this not being a `feature` spec, but I'm not exactly sure what the best practice around this would be. Thoughts?

app/controllers/employees_controller.rb
```diff
@@ -7,17 +7,11 @@ def new
   def create
     @employee = Employee.new(employee_params)
 
-    if unknown_employee(@employee.slack_username)
-      flash.now[:error] = "There is not a slack user with the username \"#{@employee.slack_username}\" in your organization."
-      render action: :new
-    elsif @employee.save
+    if @employee.save
```
jessieay: yessssss so much better

app/views/employees/_form.html.erb
jessieay: hey! yes that is weird. I just pulled down your branch and played with it for a few mins, saw same thing.

my guess is something similar to this: https://github.com/thoughtbot/factory_girl/issues/714

> webmock doesn't actually trigger until you're in a spec
 
while I'd love to get to the bottom of this, giving this a second look makes me think we should take a slightly different approach. 

rather than adding this check as an active record validation, which means it will be checked every time a user is created, maybe we can add the check as a method in `Employee` so it only gets hit when we use that method (in the controller, for instance).

My reasoning is similar to this: http://stackoverflow.com/a/4830883/1019369

Hitting an API is expensive so we don't want to do it each time we create an `employee` record. Well we do....kind of. But having this check in the AR validation means we are going to be hitting this validation even when we are creating test `employee` for totally unrelated features. 

With a method rather than the AR validation we can still have a very similar controller cleanup to what you have here. I pulled down your branch and got something working, let me know what you think:

https://github.com/18F/dolores-landingham-bot/commit/efe7847fbcad025b3d5d5643c460ed05c2b9d4d3

jbcden: Nice! I was thinking something along the same lines but wanted to hear your thoughts first. I am definitely good with that change.

jessieay:
Merged this plus my few tiny updates at https://github.com/18F/dolores-landingham-bot/commit/b660a2bc098b0bda6529cd8c7a406f0c29821d81 --thanks again for your contributions!!

PR #160, opened 2016-04-12
converting from capybara-webkit to capybara-poltergeist

### LINKED ISSUE #113: Consider replacing capybara-webkit with poltergeist
This repo is currently using v1.x of capybara-webkit, which results in the following warning when running the specs:
```
WARNING: The next major version of capybara-webkit will require at least version 5.0 of Qt. You're using version 4.8.7.
```
If you're planning on upgrading to v2 of capybara-webkit when it comes out, then it will require anyone who is contributing to this repo to have Qt 5.0 installed on their machine, which in turn requires the full Xcode app to be installed.

Having to download a multi-gigabyte app just to be able to run some tests is not a great developer experience IMO. Instead, I recommend switching to [Poltergeist](https://github.com/teampoltergeist/poltergeist), which does not depend on Xcode, and works just as well as capybara-webkit. In some ways, it's even better because it does not require `xvfb`, and makes running tests on Travis or Vagrant easier.

Here is an example of the changes required to make the switch: https://github.com/codeforamerica/ohana-web-search/pull/601/files

fentie:
* Fixes #113
* Updating gemfile
* Updated travis config
* Added new requirement to CONTRIBUTING.md
* Included new dependency in rails spec helper

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -63,6 +63,9 @@ If you have previously run a project on a different port, a `.foreman` file
 may be generated at the root of your directory. If so, make sure that this
 file is set to port `5000` or you will be unable to authenticate locally with MyUSA.
 
+### Testing
+Testing is done using `capybara-poltergeist`, which requires a local install of [phantomjs](https://github.com/jonleighton/poltergeist#installing-phantomjs).
+
 ### Required Keys
 
 The setup script creates a `.env` file with a dummy environment configuration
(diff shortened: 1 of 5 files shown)


CONTRIBUTING.md
```diff
@@ -63,6 +63,9 @@ If you have previously run a project on a different port, a `.foreman` file
 may be generated at the root of your directory. If so, make sure that this
 file is set to port `5000` or you will be unable to authenticate locally with MyUSA.
 
+### Testing
+Testing is done using `capybara-poltergeist`, which requires a local install of [phantomjs](https://github.com/jonleighton/poltergeist#installing-phantomjs).
```
jessieay: glad we are adding a section on testing to the docs! great idea. 

In addition to this line about installing poltergeist, how about adding something like:

You can run the entire test suite by running `rake`

jessieay:
one small comment, otherwise looks good to go! :hand:

jessieay:
Merged to master at b971d8f4b61fa9bd285d286f8d2e8c3f01f83ffc -- thank you!!

PR #165, opened 2016-04-30
Updating the setup script to create the tmp/pids dir

jrmils89:
Foreman tries to use the tmp/pids directory when it starts and will error
out at times if it's not there.

This just adds a small mkdir to the setup script to ensure that the directory
is created as part of the setup


afeld:
Worth looking at: https://github.com/collectiveidea/delayed_job/issues/891#issuecomment-196345863 but will merge this as a short-term fix.

PR #166, opened 2016-04-30
Fix Setup

dijonkitchen:
Update Puma gem to latest version to solve bin/setup OpenSSL issue on Mac OS X El Capitan
Update documentation for localhost port

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -57,11 +57,11 @@ is [purposefully excluded from the project's `Gemfile`][exclude].
 [foreman]: https://github.com/ddollar/foreman
 [exclude]: https://github.com/ddollar/foreman/pull/437#issuecomment-41110407
 
-The application will run locally at `http://localhost:5000/`.
+The application will run locally at `http://localhost:8000/`.
 
 If you have previously run a project on a different port, a `.foreman` file
 may be generated at the root of your directory. If so, make sure that this
-file is set to port `5000` or you will be unable to authenticate locally with MyUSA.
+file is set to port `8000` or you will be unable to authenticate locally with MyUSA.
 
 ### Testing
 Testing is done using `capybara-poltergeist`, which requires a local install of [phantomjs](https://github.com/jonleighton/poltergeist#installing-phantomjs).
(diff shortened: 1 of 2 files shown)


Gemfile.lock
```diff
@@ -159,7 +159,7 @@ GEM
       slop (~> 3.4)
     pry-rails (0.3.4)
       pry (>= 0.9.10)
-    puma (2.13.4)
+    puma (3.4.0)
```
afeld: I think this means we can undo the changes from https://github.com/18F/dolores-landingham-bot/pull/165, yeah? Also, mind dropping a link here to the issue that described the fix?

/cc @jrmils89

dijonkitchen: I still needed those changes for this fix to work.

CONTRIBUTING.md
```diff
@@ -57,11 +57,11 @@ is [purposefully excluded from the project's `Gemfile`][exclude].
 [foreman]: https://github.com/ddollar/foreman
 [exclude]: https://github.com/ddollar/foreman/pull/437#issuecomment-41110407
 
-The application will run locally at `http://localhost:5000/`.
+The application will run locally at `http://localhost:8000/`.
 
 If you have previously run a project on a different port, a `.foreman` file
 may be generated at the root of your directory. If so, make sure that this
-file is set to port `5000` or you will be unable to authenticate locally with MyUSA.
+file is set to port `8000` or you will be unable to authenticate locally with MyUSA.
```
afeld: The local port used seems to be inconsistent throughout the project (some places still say `5000`, some `8000`)...shall we pick one?

afeld:
Going to merge this to unblock people, but we should work out the port issue separately.

CONTRIBUTING.md
jessieay: I believe the default port for foreman is 5000...could be wrong. Either way, it is possible to set a default by creating a `.foreman` file -- perhaps we can do that for this project so we know people will be on the same port https://ddollar.github.io/foreman/#DEFAULT-OPTIONS

PR #167, opened 2016-04-30
Fix Port Inconsistencies

dijonkitchen:

PR #168, opened 2016-04-30
Simplify documentation

dijonkitchen:


CONTRIBUTING.md
```diff
@@ -49,15 +49,15 @@ To get started, run `bin/setup`
 
 After setting up, you can run the application using [foreman]:
 
-    % foreman start
+    foreman start
```
jessieay: hey there! I usually use a `%` of `$` to indicate that a command is on the command line (my zsh setup uses `%` as a prompt but I know `$` is more common)

Do you think having this here is confusing? how else do you think we can communicate that a blob of code is for the command line?

CONTRIBUTING.md
```diff
@@ -49,15 +49,15 @@ To get started, run `bin/setup`
...
 
 If you don't have `foreman`, see [Foreman's install instructions][foreman]. It
 is [purposefully excluded from the project's `Gemfile`][exclude].
 
 [foreman]: https://github.com/ddollar/foreman
 [exclude]: https://github.com/ddollar/foreman/pull/437#issuecomment-41110407
 
-The application will run locally at `http://localhost:5000/`.
+The application will run locally at http://localhost:5000/.
```
jessieay: this seems conflicting with https://github.com/18F/dolores-landingham-bot/pull/166/files ?

dijonkitchen: @jessieay The port change in that was reverted in #167 .

CONTRIBUTING.md
ccostino: @jessieay and at @dijonkitchen:  I've seen this come up a few times in some other conversations and the doc working group and it seems that the general consensus is to not have the prompt as it makes it easier to copy the command to run it.  Personally I like having it as it's a clear call out that it's a shell command, but what we could do instead is have it be wrapped in the code/monospace formatting (by using the back ticks) and indented (which I think it already is), like so:

    `foreman start`

And using slightly clearer language above, e.g., "After setting up, you can run the application by running the following [foreman] command in your shell:" or something to that effect.

jessieay: cool that works for me!

PR #172, opened 2016-04-30
Setup documentation modified

eaylward8:


CONTRIBUTING.md
```diff
@@ -63,6 +68,8 @@ If you have previously run a project on a different port, a `.foreman` file
 may be generated at the root of your directory. If so, make sure that this
 file is set to port `5000` or you will be unable to authenticate locally with MyUSA.
 
+If your server isn't defaulting to Port 5000, you may have to add a .foreman file to root directory. In the file, add "port: 5000".
```
jessieay: 👍

CONTRIBUTING.md
```diff
@@ -113,6 +120,7 @@ development with the following:
 4. Edit the `AUTH_DOMAIN` value in your local `.env` file such that the `is_permitted` method in
    `/app/controllers/auth_controller` will accept the email address you used in your MyUSA
    application.
+  For example, if you use gmail, change the AUTH_DOMAIN to "gmail.com"
```
jessieay: 👍

CONTRIBUTING.md
```diff
@@ -124,6 +132,12 @@ development with the following:
   end
 ```
 
+## Granting Yourself Admin Access
+
+Once you've created an account with MyUSA, open the Rails console and change your admin status to "true". This will allow you to create and schedule messages.
+
+Note: scheduled messages may not send if you're using over the weekend, which is due to the business_time gem referenced above.
```
jessieay: WOW this is important. Wonder if we should be part of a separate bullet point so it is highlighted even more?

CONTRIBUTING.md
```diff
@@ -45,6 +45,11 @@ What constitutes a business day is managed by the gem `business_time` and is con
 
 ### App setup
 
+Before running bin/setup, ensure that 'foreman' is removed from the Gemfile.
```
jessieay: ooooh what happens if it is already installed? Maybe we should update setup script to play more nicely with the Gemfile as is...

CONTRIBUTING.md
```diff
@@ -45,6 +45,11 @@ What constitutes a business day is managed by the gem `business_time` and is con
 
 ### App setup
 
+Before running bin/setup, ensure that 'foreman' is removed from the Gemfile.  
+
+Prior to running 'bin/setup', it may be necessary to launch the postgres server using 
+```postgres -D /usr/local/var/postgres``` You can also use the OS X app.
```
jessieay: good call - this can trip ppl up. Perhaps we should link to the postgres app?

jessieay:
Hi @eaylward8 - thanks so much for submitting this doc improvements! I left a few comments / suggestions. Do you have time to address those? If not, I can do them myself, but wanted to give you the opp since you are the original PR author.

Thanks again!!

🌈

eaylward8:
Hey @jessieay - Sorry about that, it fell off my radar! I'll take a look today.  I was working with some other people at a meetup in NYC, so I'll do my best to remember everything we did.

CONTRIBUTING.md
eaylward8: I'd say this deserves to be highlighted one way or another, because you can't do much without being an admin.  Not sure if there's a better way of getting admin access or if the way we did it (as described here) is appropriate.

CONTRIBUTING.md
eaylward8: This part of the setup confused us.  There's a line in the App Setup section that says:

> "If you don't have foreman, see Foreman's install instructions. It is _purposefully excluded_ from the project's Gemfile." 

It says that the Foreman gem is excluded, but it's not - it's actually still in the Gemfile (line 10).  This seemingly conflicting information confused us during the setup process (we are all pretty inexperienced devs).  Again, I'll have to defer to you on how to best address this issue - all I can really do is say that this was a pain point for us during setup.

CONTRIBUTING.md
eaylward8: A link and some explicit instructions would be great, especially for newbies like us :)

eaylward8:
@jessieay - I left some comments for you, sorry it took me until now to get to it.  Any questions, just let me know.  Hope it helps!

CONTRIBUTING.md
jessieay: oh yeah...that is totally my fault. Usually I use foreman locally only so I do not include it in the Gemfile. This is because it's usually a best practice to leave development dependencies that are not required out of the project. eg: I might like foreman but maybe you just want to use the regular rails server.

In this project, however, we are using foreman in all envs so we need it. Let's merge this as-is and I will open a separate Pr to deal with foreman instructions.

jessieay:
Merged to master at 924787f66c51a9510bbd7af7ce1dd3c4596e4685 -- Thank you!!! 🌈

Issue #157, opened 2016-03-16
Username validation or send failure error


nicoleslaw:
I'm not sure how hard this would be, but Dolores should let us know if a username is failing to receive messages. I added @amandaschonfeld's initial username when she joined Slack, and then she changed it—I didn't know the messages weren't coming through until she DMed me asking what happened and I realized the name change wasn't throwing an error etc.

jessieay:
Steps for completing this:

1) Figure out if we can update an employee's slackusername in our system with it is changed on slack

if above is not possible: 
1) Figure out what error is thrown when we attempt to send a slack message to an invalid username
2) Surface that error to app admins -- perhaps via a daily email? Or maybe we should just set up newrelic instead and let that notify us? https://github.com/18F/dolores-landingham-bot/issues/21

jrmils89:
Hey all. So I took an approach here that it would probably be better to send out slack messages using the slack user id (which shouldn't change when a username is change) instead of raising errors on bad slack usernames. I started working on this and realized it's a fairly substantial change.

I think I covered everything, and it's all housed here: https://github.com/jrmils89/dolores-landingham-bot/commits/use-slack-user-ids. I know this wasn't the exact purpose of this issue, so I thought it'd be a good idea to add a comment here, and you guys could let me know how you'd like me to proceed (pull request, make changes, do nothing since it's out of the scope of the ticket). If anything, it's an interesting exercise and hopefully saved some people time from doing this research themselves. 

I would have liked to test this with a larger dev DB to test backwards compatibility, but it should work going backwards (in theory).

A small note. Messages seem to actually be posted using the channel id as opposed to the username. It seems to me what likely happened here is that between being added to dolores and the first message being sent, the username was changed. Now the user id is being set on save, and the channel id is set based on the slack user id. It would probably be pretty simple to (now that the user id is saved on creating the user) to check for username updates when an email gets sent (or daily, or whenever makes sense).

jessieay:
@jrmils89 sendings message via Slack ID instead of username is a GREAT idea. We're still gonna want to add users via Slack username since that is more user-friendly, but I think using something that is unique and unchanging is super ideal. Love that idea.

Depending on how much work it is, I'd suggest you do this in 2 steps:

1. Save slack user ids and use those to send messages
[...]

PR #174, opened 2016-05-06
Use slack user ids

jrmils89:
Fixes issue(s) #157 

- Save slack user id when the employee is created
- Check for missing slack user ids when sending a message, and add them if it can
- Look up the channel id (used to post messages) by slack user id instead of slack username
- Log sent message errors if slack channel for a username can't be found (indicating a bad username) at some point along the way
- DB migration and test updates

@jessieay @nicoleslaw

### DIFF
--- a/app/controllers/employees_controller.rb
+++ b/app/controllers/employees_controller.rb
@@ -35,6 +35,7 @@ class EmployeesController < ApplicationController
     @employee = Employee.find(params[:id])
     @employee.slack_username = params[:employee][:slack_username]
     @employee.validate_slack_username_in_org
+    @employee.add_slack_user_id_to_employee
 
 
     if @employee.errors.full_messages.empty? && @employee.update(employee_params)
--- a/app/controllers/employees_controller.rb
+++ b/app/controllers/employees_controller.rb
@@ -7,6 +7,8 @@ class EmployeesController < ApplicationController
   def create
     @employee = Employee.new(employee_params)
     @employee.validate_slack_username_in_org
+    @employee.add_slack_user_id_to_employee
+
 
     if @employee.errors.full_messages.empty? && @employee.save
       flash[:notice] = "Thanks for adding #{@employee.slack_username}"
--- a/app/models/employee.rb
+++ b/app/models/employee.rb
@@ -29,7 +29,7 @@ class Employee < ActiveRecord::Base
 
   def add_slack_user_id_to_employee
     user_id = EmployeeFinder.new(slack_username).slack_user_id
-    if !user_id.nil?
+    if user_id.present?
       self.slack_user_id = user_id
     end
   end
--- a/app/models/employee.rb
+++ b/app/models/employee.rb
@@ -28,8 +28,9 @@ class Employee < ActiveRecord::Base
   end
 
   def add_slack_user_id_to_employee
-    if !EmployeeFinder.new(slack_username).slack_user_id.nil?
-      self.slack_user_id = EmployeeFinder.new(slack_username).slack_user_id
+    user_id = EmployeeFinder.new(slack_username).slack_user_id
+    if !user_id.nil?
+      self.slack_user_id = user_id
     end
   end
 end
--- a/app/models/employee.rb
+++ b/app/models/employee.rb
@@ -26,4 +26,10 @@ class Employee < ActiveRecord::Base
       )
     end
   end
+
+  def add_slack_user_id_to_employee
+    if !EmployeeFinder.new(slack_username).slack_user_id.nil?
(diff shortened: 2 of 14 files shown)


app/services/message_sender.rb
```diff
@@ -9,13 +9,20 @@ def initialize(employee, message)
...
+
     if employee.slack_channel_id.nil?
-      channel_id = SlackChannelIdFinder.new(employee.slack_username, client).run
+      channel_id = SlackChannelIdFinder.new(employee.slack_user_id, client).run
       employee.slack_channel_id = channel_id
       employee.save
     end
 
-    if employee.slack_channel_id
+    if !employee.slack_channel_id.nil?
```
jessieay: how about simplifying to `if employee.slack_channel_id.present?` ?

app/services/slack_user_finder.rb
```diff
@@ -3,6 +3,12 @@ def existing_user?
     !slack_user.nil?
   end
 
+  def user_id
+    if !slack_user.nil?
```
jessieay: what about `if slack_user.present?` ?

db/migrate/20160430175239_add_slack_user_id_to_employee.rb
```diff
@@ -0,0 +1,5 @@
+class AddSlackUserIdToEmployee < ActiveRecord::Migration
+  def change
+    add_column :employees, :slack_user_id, :string
```
jessieay: nice! can we add an index on this too? Looks like we're querying on it.

app/models/employee.rb
```diff
@@ -26,4 +26,11 @@ def validate_slack_username_in_org
       )
     end
   end
+
+  def add_slack_user_id_to_employee
+    user_id = EmployeeFinder.new(slack_username).slack_user_id
+    if !user_id.nil?
```
jessieay: how about ` if user_id.present?`

app/controllers/employees_controller.rb
```diff
@@ -33,6 +35,7 @@ def update
     @employee = Employee.find(params[:id])
     @employee.slack_username = params[:employee][:slack_username]
     @employee.validate_slack_username_in_org
+    @employee.add_slack_user_id_to_employee
```
jessieay: wondering if we should move some of this logic to its own class. Feels like the controller is getting bigger and bigger. Doesn't need to happen now, but if you were up for it that would be cool.

app/services/message_sender.rb
```diff
@@ -9,13 +9,20 @@ def initialize(employee, message)
   def run
     configure_slack
 
+    user_id = EmployeeFinder.new(employee.slack_username).slack_user_id
+
+    if employee.slack_user_id.nil? && !user_id.nil?
```
jessieay: what about: `if employee.slack_user_id.nil? && user_id.present?`

app/services/message_sender.rb
```diff
@@ -9,13 +9,20 @@ def initialize(employee, message)
   def run
     configure_slack
 
+    user_id = EmployeeFinder.new(employee.slack_username).slack_user_id
```
jessieay: maybe we should call this var `slack_user_id` for clarity

app/services/message_sender.rb
```diff
@@ -30,7 +37,14 @@ def run
           error: error,
         )
       end
+    else
+      create_sent_scheduled_message(
+        employee: employee,
+        scheduled_message: message,
+        error: StandardError.new("Was unable to find a slack channel for this username"),
```
jessieay: oh this is an interesting way to surface this error. very nice! Wondering if we should be more specific with the error message. Something like: `"unable to find slack user with name #{employee.slack_username} and slack user id #{employee.slack_user_id}"` -- I think having that info in the error msg would be very helpful in debugging!

jessieay:
This looks awesome - just a few small comments from me. Once they are addressed, looks good for squash n merge!

app/controllers/employees_controller.rb
jrmils89: Yeah, I agree that it's getting a little bloated. I could make a new issue for this if you want? I'm not sure I have a ton of time to get to this particularly soon

jrmils89:
OK, this should be ready to go now (after I forgot to push the error text test change the first time). Made the changes based on the feedback here (everything except for abstracting out the controller logic...new issue maybe?).

jessieay:
awesome, thanks for making updates! Merged at de8cec44ec78e128fb7c602813a62651f8e303d3

I will try to tackle some controller refactors in near future. Appreciate this elegant solution!! 🌈

PR #177, opened 2016-05-17
Upgrade to Ruby 2.3.1

jessieay:
* Was getting buildpack error for previous version

screen shot 2016-05-17 at 10 07 14 am

cc @gemfarmer @ccostino


ccostino:
Hey @jessieay!  Would it be worth just upgrading to 2.3.1 instead since that is also out now?

jessieay:
@ccostino good call, didn't even realize it was out! updated.

ccostino:
Thanks, @jessieay! :-)

PR #178, opened 2016-05-17
Validate uniqueness of Employee slack username with scope

jessieay:
Fixes issue: was not able to delete an employee and re-add
* We are using paranoia to soft delete
* Need to scope uniqueness validations to records that are not deleted

### DIFF
--- a/app/models/employee.rb
+++ b/app/models/employee.rb
@@ -3,10 +3,16 @@ class Employee < ActiveRecord::Base
 
   has_many :sent_scheduled_messages, dependent: :destroy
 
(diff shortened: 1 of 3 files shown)


jessieay:
ping ping @ccostino @gemfarmer

ccostino:
Hi @jessieay!  I'm sorry for the delay in taking a look at this.  I just checked this over and all looks good to me, great catch in accounting for our use of paranoia.  Thank you!

jessieay:
thanks for the review, @ccostino !

PR #184, opened 2016-06-17
Upgrade Ruby, libs

jessieay:
* No more outdated libs


jessieay:
Self-merging since this is a pure upgrade PR

PR #186, opened 2016-06-24
Add rubocop and code climate

jessieay:
cc @gemfarmer @ccostino

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -1,0 +1,14 @@
+engines:
+  rubocop:
+    enabled: true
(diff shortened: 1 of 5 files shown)


ccostino:
Nice, @jessieay!  Looks like codeclimate already found several things.  I don't have access to the details though, I was curious what it found.

jessieay:
@ccostino oh weird, when you click it doesn't let you see analysis? let me see if I can change permissions

ccostino:
There we go!  Now I see the details.  Thank you!

Issue #180, opened 2016-05-25
Handle regular policy reminders


gemfarmer:
This is here as a reminder to investigate further, not as an actionable ticket.

From @batemapf via slack
> hey folks - i am interested in automating some of the regular policy reminders that ops and other teams need to send out. for instance, we manually send quarterly reminders about not starting work for clients until we have a signed agreement. would be great to have this delivered automagically to each individual at the best possible time (i.e. taking into account time zones, work patterns). extra points if the recipient can confirm they understand or raise a hand if they don't / have a question. any suggestions on the above?

@ccostino:
> Oh interesting, that would essentially be adding functionality to @mrs_landingham to send repeat messages at set intervals.

me:
> there is also an issue already to handle two way communication. (edited)
> This would be a little odd because it pertains to a particular message itself instead of being a stand alone question.

jessieay:
This seems right in line with Mrs. Landingham's core purpose!

Implementation brainstorm:

Add a new `message_type` column to `scheduled_messages` -- if it is an `onboarding` message it gets sent once. If it is `quarterly` then we send it out on the first day of each quarter.

jessieay:
### Requirements for this feature:

#### Description:

Given that I am an admin
And I am logged in
When I visit the new_scheduled_messages path 
And I add a title and message body to a scheduled message
And select 'quarterly' as the message type
Then the 'Business days after employee starts to send message' and 'Time of day to send message (uses employee's local time zone)' fields cannot be filled out 
And the message should be sent on October 1, January 1, April 1 and July 1
If any if the above dates is on a weekend, it should send the message on the following Monday

#### Testing requirements

At least one capybara end-to-end test for the above functionality (see `spec/features` directory for other capybara tests)

Unit tests for any new underlying functionality (RSpec)

### Suggested implementation (not required)

We are already using the `holidays` gem, which you can use to make sure messages don't get sent out on a weekend https://github.com/18F/dolores-landingham-bot/blob/master/Gemfile#L25

Regular messages are for onboarding purposes and are only sent once. Now that we are introducing a new type of message. I would call the new column on `scheduled_messages` `type` have it be an ActiveRecord enum that defaults to `onboarding` but can also be set to `quarterly` (like in the case for this feature)

Once `quarterly` is selected, the other fields that are specific to onboarding-type messages ( 'Business days after employee starts to send message' and 'Time of day to send message (uses employee's local time zone)')  can either be hidden or disabled. whatever is easier.

[...]

PR #187, opened 2016-06-24
Add quarterly reminders feature

danmalks:
Fixes issue #180

This feature looked simple, but required changes that touched a few
different classes.  It was hard not to design something that could
simply send messages on specific dates (instead of being constrained
to the four quarter dates), but what is here can be easily adapted
to allow for scheduling messages on any date.

I tried to leave the existing models and service objects alone
as much as I could.  Now there are two types of messages: "onboarding"
which are those sent at intervals from an employee's start date,
and "quarterly" which are sent four times a year on specific dates.

Quarterly messages will repeat year after year, this was not
addressed in the requirements so I made an assumption.

Also a quarterly message will be sent to a new employee up to
one week after the quarterly message date; this was also an
assumption and easily undone or adjusted.

Some may object to the helper methods I scatter at the bottom
of my specs.  Some of them are repeated in different files.
This is no accident--I prefer to have, in a spec and in a test
example, all the contextual information at hand to describe
what is happening in the tests.  Then you can look at any
test and know from A to Z what is happening, without needing
to hunt for mystery values hidden in helpers, etc.

This is an alpha commit, I suppose.  feedback is welcome.


mtorres253:
wooo! Thanks @danmalks! @jessieay let us know what you think

jessieay:
Hey @danmalks, excited to dig into code review this afternoon.

In the meantime, I just added code climate to this repo to address any style things that we should change. For example, I noticed a large inline comment but I personally don't use inline comments unless it is a very particular situation. 

Anyway, I am happy to leave style-related comments but if you rebase from master you'll be able to see the code climate report on this PR which might be a more efficient way to address style issues. :)

app/models/scheduled_message.rb
```diff
@@ -5,11 +5,13 @@ class ScheduledMessage < ActiveRecord::Base
   has_many :sent_scheduled_messages, dependent: :destroy
 
   validates :body, presence: true
-  validates :days_after_start, presence: true
+  validates :days_after_start, presence: true, unless: "quarterly?"
```
jessieay: nice! can we write a small unit test for this validation?

spec/models/sent_scheduled_message_spec.rb
```diff
@@ -35,4 +35,15 @@
   describe "Delegated methods" do
     it { should delegate_method(:slack_username).to(:employee) }
   end
+
+  describe "Scopes" do
+    describe "self.by_year" do
```
jessieay: very nice! this is super small but I usually put `.by_year` in the description for class methods (if it were an instance method, I'd do `#by_year`)

spec/services/quarterly_message_integration_spec.rb
```diff
@@ -0,0 +1,57 @@
+require "rails_helper"
+require "business_time"
+
+# I really hate putting this test here as it does not correspond
```
jessieay: I agree that this is faster than a feature spec but I propose we do one of the following:

1. create a class that encapsulates this behavior so we can test that class in order to ensure that everything is working together, like you say
OR
2. just make it a feature spec. would be slower (like you say), but having this type of random integration spec floating around makes it harder for future developers to know what types of tests they need.

db/migrate/20160623213452_remove_null_constraint_on_days_after_start.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveNullConstraintOnDaysAfterStart < ActiveRecord::Migration
+  def change
+    change_column_null(:scheduled_messages, :days_after_start, true)
```
jessieay: love `change_column_null` ! just making sure: this is reversible, right?

app/views/scheduled_messages/_form.html.erb
```diff
@@ -4,6 +4,13 @@
   <%= form.input :body, label: "Message body",
     hint: "Don't forget, you can format your message using #{link_to "Slack's message formatting rules",
     "https://slack.zendesk.com/hc/en-us/articles/202288908-Formatting-your-messages", target: "_blank"}.".html_safe %>
+  <%= check_box_tag :quarterly, checked: false %> Quarterly
+  <div style="display:none" class="quarterly_message_explanation">
+    This message will be sent to all employees quarterly
+    <br />
```
jessieay: any way to do this without `br` tags? I am not a frontend pro but I know they are generally not favored...

app/services/quarterly_message_employee_matcher.rb
```diff
@@ -0,0 +1,52 @@
...
+    message_time_value = TIME_OF_DAY_TO_SEND
+    employee_current_time_value >= message_time_value
+  end
+
+  def current_year
+    Time.now.year
+  end
+
+
+
```
jessieay: can we clean up these extra newlines?

app/services/quarterly_message_employee_matcher.rb
```diff
@@ -0,0 +1,52 @@
+require "business_time"
+
+class QuarterlyMessageEmployeeMatcher
```
jessieay: nice! I love the consistency of this class with our other matcher class...wondering if we should rename the other one now to clarify difference (eg: `OnboadingMessageMatcher`

jessieay: oh wait nevermind, I see now that the other class calls this one.

app/services/next_working_day_finder.rb
```diff
@@ -0,0 +1,20 @@
+require "business_time"
+
+class NextWorkingDayFinder
+
+  def self.run(date = Time.today)
+    self.new(date).run
+  end
+
+  def initialize(date = Time.today)
+    @date = date
```
jessieay: let's create an `attr_reader` for date. that's a style we've been using in the other service classes.

app/services/quarterly_message_day_verifier.rb
```diff
@@ -0,0 +1,37 @@
+require "business_time"
+
+class QuarterlyMessageDayVerifier
+
+  DEFAULT_MONTH_DAYS_TO_SEND_ON = [ [1,1], [4,1], [10,1], [7,1] ]
+
+  def initialize(date: Time.today)
+    @date = date
+    @current_year = @date.year
```
jessieay: att_readers for these two would be good

app/services/quarterly_message_day_verifier.rb
```diff
@@ -0,0 +1,37 @@
+require "business_time"
+
+class QuarterlyMessageDayVerifier
+
+  DEFAULT_MONTH_DAYS_TO_SEND_ON = [ [1,1], [4,1], [10,1], [7,1] ]
```
jessieay: nice!

jessieay:
Left a few comments. There are some other minor stylistic things but if you rebase you will catch those via rubocop.

I would like to see 1-2 feature specs for this using capybara. That would cover the javascript stuff that happens in the form (test that confirms desired fields are on the other page) and that an admin user can create a quarterly scheduled message via the UI.

Just let me know if you have any questions!

app/models/scheduled_message.rb
pdbradley: good catch @jessieay will do

spec/models/sent_scheduled_message_spec.rb
pdbradley: yes, that's my custom too.  whoops will fix

spec/services/quarterly_message_integration_spec.rb
pdbradley: @jessieay good point on number 2.  I didn't think about that. Is it permissible in a feature test here to create a bunch of employees via factory?  The reminder can still be created via capybara.  Some folks are sticklers for feature tests being 100% point and clicky and not using any behind the scenes help.

jessieay: oh yes totally fine with me to create seed data in feature specs. If you look at our existing feature specs, we are doing just that: https://github.com/18F/dolores-landingham-bot/blob/master/spec/features/edit_scheduled_messages_spec.rb

Now that I rethink this, though, it seems like what we're really testing here is that `DailyMessageSender.new.run` is doing its job. 

Hard to test clockwork job since that is not a feature with a UI. So maybe capybaras is not the anwswer.

Thoughts on moving all this to `DailyMessageSender.`  unit tests? I do like your usage of Timecop here...

app/services/quarterly_message_employee_matcher.rb
pdbradley: @jessieay yeah the other class screams for renaming just as you said.  I just wasn't sure how many fingerprints I was allowed to leave on the code so I tried not to go overboard.

pdbradley:
@jessieay there is one feature test, spec/features/create_quarterly_message_spec.rb, I guess that isn't what you mean?  It does use the javascript.

spec/services/quarterly_message_integration_spec.rb
pdbradley: yes, I felt like it sort of would fit into DailyMessageSender also.  Or, we can just delete it altogether; sometimes when things don't fit, it means...they don't fit.  :)

jessieay:
@pdbradley ah yes I missed that feature spec! Apologies. I will say that I'd probably remove the expectations around date (`expect Class.count to eq(1)`) from those to make them more "feature"-y, but no biggie.

app/services/quarterly_message_employee_matcher.rb
jessieay: yes always a tricky balance to make PRs small and clean up things as you go along! if you think of a good class name change, by all means go for it. If not then no worries.

db/migrate/20160623213452_remove_null_constraint_on_days_after_start.rb
pdbradley: @jessieay yep it is reversible.

pdbradley:
@jessieay I think you should disable the documentation check on code climate.  That's a little bit over the top, to me, to require comments over each class heading.  I didn't want to conform to it until I was sure you were serious about that.  I think you dropped CC into the project without much tweaking of the config.

jessieay:
@pdbradley oh yes, 100% agree on that! Was probably not a great idea to introduce code climate before I myself had done a PR using the rubocop file that I added. Am using settings that I've used before (via https://github.com/thoughtbot/guides/tree/master/style/ruby) so thought there wouldn't be any surprises, but I did not expect this to have the class-comment rule. Def don't include those and sorry for any confusion!!

pdbradley:
@jessieay this is our first microbid so I'm not sure at what point this can be deemed as 'passing acceptance criteria'.  However, if we can get it to that point pretty soon that would be great.; I'm traveling to Europe tomorrow and will have intermittent internet access for about 20 hours.

jessieay:
@danmalks this all looks good to me and fulfills the auction requirements. 

If there are further refinements we need I can take care of them myself in a separate PR.

Travel safely and we will be in touch!

danmalks:
@jessieay thanks so much. pleasure for us to be a part of this development effort and we look forward to working together in the future.

jessieay:
Merged to master at 4de847866f9d56baed7aea4ea31048ff335fb13a -- thank you!!

PR #191, opened 2016-07-13
Adds dev configuration in vagrant.

dsummersl:
This PR adds a vagrant configuration that installs all dependencies required to get a dev instance up and running. One can:

 1. clone the repository
 2. run `vagrant up` to bring up a new linux server on their local dev box.
 3. login via `vagrant ssh` to run tests, foreman, etc (eg `cd /vagrant; foreman start`)


jessieay:
Hi @dsummersl - thanks for the PR! I myself do not have experience w/Vagrant but this seems like a good thing to have so our contributors can get up and running more quickly.

Can you add a section / some notes on Vagrant to our contributing docs? That will ensure that people know we have this: https://github.com/18F/dolores-landingham-bot/blob/master/CONTRIBUTING.md

Thanks!

dsummersl:
@jessieay I've added a short primer for its use.  I'm seeing the codeclimate failure, but I'm not sure what to do about it...any ideas?

PR #162, opened 2016-04-13
[WIP] Connects #72:  Adds employee import functionality

ccostino:
This changeset adds an importer to retrieve all Slack users within the organization.  It is configured as a Rake task and includes support for performing a dry run of the import.

You can run the import task with the following command:

```
rake employees:import
```

You can also perform a dry run of the import to see if there will be any errors:

```
rake employees:import:dry_run
```

I brought this branch up-to-date with the most recent activity on the `master` branch and made sure all of the tests passed, but it has been several months since I last worked on this and I am unsure if it still meets the needs of #72.  It does pull in all available Slack usernames in the organization and defaults the start date to whenever the import was run.

In the future, it'd be nice to leverage the Team API for this information instead as it should also contain the employee start dates.  I have added several `TODO` notes in the code around this.

If someone is able to take a look at this and carry the work over the finish line (the branch - `cc-72-import-employees` - is pushed remotely so you can get to the code easily), I'd greatly appreciate it!  Please feel free to reach out to me with any questions/concerns/feedback regarding what I've done so far.

/cc @jessieay and @gemfarmer

Issue #72, opened 2015-10-16
Add ability to retrieve all users/employees in an organization from Slack


ccostino:
When the Dolores app first starts, she starts with an empty list of employees.  It would be very helpful if we could provide some way of retrieving all employees and placing them into the app automatically for the admins to make things easier.

jessieay:
@ccostino we can do that with the Slack API, but we would not have the start date so would want to do it in such a way that Dolores admins know they have to add start date (Slack API does not make sign up date available, if my memory serves me)

melodykramer:
As an interim solution, even just a dropdown with everyone's Slack handles would be helpful. My current workflow is: looking to see who wants to be added, finding their Slack handle, typing it in. Not everyone's handle is intuitive, but a drop-down menu containing all 18F Slack handles would save the step, and then the start date could be entered manually.

jessieay:
@melodykramer that's a great idea and wouldn't be too difficult to implement

melodykramer:
Awesome, thanks @jessieay!

jessieay:
@ccostino seems like you were working on this -- any update? should we unassign you to give others the opp to work on it?

ccostino:
@jessieay:  Yikes, it's been a while and I honestly can't remember where I left off.  It doesn't look like I submitted a PR for this yet but in looking at the work it appears like I had gotten pretty far.  I can get the branch up-to-date and push it remotely and/or make a PR out of it for someone else to take a look and pick up as I am pretty well booked at this point.

I'm more than happy to let someone else either wrap it up or take a crack at it and can try and bring them up-to-speed with what I have so far if it would help.

jessieay:
# Auction requirements

One feature-level test (Capybara) to confirm the following behavior:

Given that I am in the slack user in the same team as the bot user
And there is not an employee record for me in the Dolores system
Then my slack username and id should be imported into the Dolores system as an employee record once per day at 3am UTC

Any other classes/methods that are introduced in order to complete the above feature should have unit tests. 

# Other helpful data (not required but will get you started)

* An unfinished WIP version of this feature exists at https://github.com/18F/dolores-landingham-bot/pull/162/files -- feel free to use any/all of that code to get started.

* We already have a method that retrieves all slack users in the org for the bot user: https://github.com/18F/dolores-landingham-bot/blob/master/app/services/slack_api_wrapper.rb#L23. API docs for endpoint are at https://api.slack.com/methods/users.list

[...]

PR #192, opened 2016-07-18
[Connects #72] Add nightly tasks to import new employees and update usernames

mrgilman:
I started with @ccostino's work in progress (#162), fixed the specs, added a cron task, and made a small change to persist the `slack_user_id` in addition to the username when an employee is imported.

I added a new task to update employee usernames when they are changed. This code is dependent on `slack_user_id` being populated for all employees. I saw that the field is optional in the database; it may be a good idea to run a script to populate the `slack_user_id` field for all employees and then make it `null: false`. Alternatively, I can make a change to run the updater selectively on Employee records that have the `slack_user_id` column populated.

### DIFF
--- a/app/services/employee_importer.rb
+++ b/app/services/employee_importer.rb
@@ -1,28 +1,30 @@
 require "slack-ruby-client"
 
 class EmployeeImporter
-  def initialize
-    configure_slack
-  end
+  def run(dry_run = false)
+    import_results = { created: 0, skipped: 0, dry_run: false }
+    slack_user_finder = SlackUserFinder.new("", client)
+    total_employees = slack_user_finder.users_list.count
 
-  def import(dry_run = false)
-    import_results = {created: 0, skipped: 0, dry_run: false}
-    slack_user_importer = SlackUserImporter.new("", client)
-    total_employees = slack_user_importer.slack_usernames.count
-
-    slack_user_importer.slack_usernames.each_with_index do |slack_username, index|
-      if !dry_run
-        success = import_employee(slack_username)
-      else
-        success = dry_import_employee(slack_username)
+    slack_user_finder.users_list.each_with_index do |user, index|
+      if dry_run
+        success = dry_import_employee(
+          slack_username: user["name"],
+          slack_user_id: user["id"],
+        )
         import_results[:dry_run] = true
+      else
+        success = import_employee(
+          slack_username: user["name"],
+          slack_user_id: user["id"],
+        )
       end
 
       if success
-        Rails.logger.info "#{index + 1}/#{total_employees}: Created #{slack_username}#{" (dry run)" if dry_run}".green
+        log_success(index, total_employees, user, dry_run)
         import_results[:created] += 1
       else
-        Rails.logger.info "#{index + 1}/#{total_employees}: Skipped #{slack_username}#{" (dry run)" if dry_run}".yellow
+        log_failure(index, total_employees, user, dry_run)
         import_results[:skipped] += 1
       end
     end
@@ -36,26 +38,40 @@ class EmployeeImporter
     @client ||= Slack::Web::Client.new
   end
 
-  def configure_slack
-    Slack.configure do |config|
-      config.token = ENV['SLACK_API_TOKEN']
-    end
+  def dry_import_employee(slack_username:, slack_user_id:)
+    Employee.where(
+      slack_username: slack_username,
+      slack_user_id: slack_user_id,
+    ).empty?
   end
 
-  def dry_import_employee(slack_username)
-    Employee.where(slack_username: slack_username).size == 0
-  end
+  def import_employee(slack_username:, slack_user_id:)
+    employee = Employee.new(
+      slack_username: slack_username,
+      slack_user_id: slack_user_id,
+      started_on: Time.current,
+    )
 
-  def import_employee(slack_username)
-    # TODO:  Once we can hook up the Team API, we'll want to grab the start
-    # TODO:  date of an employee from there instead of defaulting to the
-    # TODO:  current date and time.
-
-    # TODO:  We'll also want to verify that the Slack user is an actual person
-    # TODO:  Against the Team API - Slack alone cannot tell us if they are real
-    # TODO:  or not.
-
-    employee = Employee.new(slack_username: slack_username, started_on: Time.now)
     employee.save
+  end
+
+  def log_success(index, total_employees, user, dry_run)
+    info = "#{index + 1}/#{total_employees}: Created #{user['name']}"
+
+    if dry_run
+      info.concat(" (dry run)")
+    end
+
+    Rails.logger.info info.green
+  end
+
+  def log_failure(index, total_employees, user, dry_run)
+    info = "#{index + 1}/#{total_employees}: Skipped #{user['name']}"
+
+    if dry_run
+      info.concat(" (dry run)")
+    end
+
+    Rails.logger.info info.yellow
   end
 end
--- a/app/services/employee_importer.rb
+++ b/app/services/employee_importer.rb
@@ -1,0 +1,61 @@
+require "slack-ruby-client"
+
+class EmployeeImporter
+  def initialize
+    configure_slack
+  end
+
+  def import(dry_run = false)
+    import_results = {created: 0, skipped: 0, dry_run: false}
+    slack_user_importer = SlackUserImporter.new("", client)
+    total_employees = slack_user_importer.slack_usernames.count
+
+    slack_user_importer.slack_usernames.each_with_index do |slack_username, index|
+      if !dry_run
+        success = import_employee(slack_username)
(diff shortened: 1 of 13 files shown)


mrgilman:
I'm looking through the Code Climate issues. It doesn't look like the frozen string literal comment is in use anywhere else in the app. Should I add it to the files I touched, or ignore those issues?

jessieay:
Hello! re: Code Climate -- yes I would ignore the frozen string literal comment stuff. I thought I'd removed that from CC! Is your branch up to date with master? (I'd assume that it is, but just want to make sure before investigating further.

mrgilman:
Yup, it's up to date.

jessieay:
@mrgilman ok I will investigate separately. Sorry for any weird rubocop-related things. I added it pretty recently so I am still figuring out which rules I need to tweak. Just lmk if there is any other weirdness! Will take a look at the PR later this afternoon

mrgilman:
Sounds good, thanks @jessieay!

app/services/employee_importer.rb
```diff
@@ -0,0 +1,74 @@
+require "slack-ruby-client"
+
+class EmployeeImporter
+  def initialize
+    configure_slack
```
jessieay: not opposed to having this logic in two places, but wonder if we should consider putting the Slack config stuff into one place so `EmployeeImporter` and `EmployeeFinder` can share it...

app/services/employee_importer.rb
```diff
@@ -0,0 +1,74 @@
...
+  end
+
+  def import(dry_run = false)
+    import_results = {created: 0, skipped: 0, dry_run: false}
+    slack_user_finder = SlackUserFinder.new("", client)
+    total_employees = slack_user_finder.users_list.count
+
+    slack_user_finder.users_list.each_with_index do |user, index|
+      if !dry_run
```
jessieay: might be slightly more readable to have the conditional logic start with`if dry_run`

app/services/employee_importer.rb
```diff
@@ -0,0 +1,74 @@
...
+
+  def dry_import_employee(slack_username:, slack_user_id:)
+    Employee.where(
+      slack_username: slack_username,
+      slack_user_id: slack_user_id,
+    ).size == 0
+  end
+
+  def import_employee(slack_username:, slack_user_id:)
+    # TODO:  Once we can hook up the Team API, we'll want to grab the start
```
jessieay: let's remove these TODOs :)

app/services/employee_importer.rb
```diff
@@ -0,0 +1,74 @@
...
+    # TODO:  current date and time.
+
+    # TODO:  We'll also want to verify that the Slack user is an actual person
+    # TODO:  Against the Team API - Slack alone cannot tell us if they are real
+    # TODO:  or not.
+
+    employee = Employee.new(
+      slack_username: slack_username,
+      slack_user_id: slack_user_id,
+      started_on: Time.now
```
jessieay: `Time.current` ? (I always default to that in case timezones matter)

app/services/employee_updater.rb
```diff
@@ -0,0 +1,7 @@
+class EmployeeUpdater
+  def update
+    Employee.find_each do |employee|
```
jessieay: niiiice -- I forgot about `find_each` !

app/services/slack_username_updater.rb
```diff
@@ -0,0 +1,31 @@
+class SlackUsernameUpdater < SlackApiWrapper
+  def initialize(employee)
+    @employee = employee
+    configure_slack
```
jessieay: same comment as above -- any way we can put the slack ruby client config in one place?

app/services/employee_importer.rb
```diff
@@ -0,0 +1,74 @@
+require "slack-ruby-client"
+
+class EmployeeImporter
+  def initialize
+    configure_slack
+  end
+
+  def import(dry_run = false)
+    import_results = {created: 0, skipped: 0, dry_run: false}
```
jessieay: is our rubocop config enforcing no spaces between curly braces? I thought I'd copied over the tbot guide but maybe something went wrong...

app/services/employee_importer.rb
```diff
@@ -0,0 +1,74 @@
...
+      else
+        success = dry_import_employee(
+          slack_username: user["name"],
+          slack_user_id: user["id"]
+        )
+        import_results[:dry_run] = true
+      end
+
+      if success
+        Rails.logger.info "#{index + 1}/#{total_employees}: Created #{user["name"]}#{" (dry run)" if dry_run}".green
```
jessieay: these statements might be more readable if wrapped into well-named private methods - eg `log_success` or something

app/services/employee_importer.rb
```diff
@@ -0,0 +1,74 @@
...
+  def import_employee(slack_username:, slack_user_id:)
+    # TODO:  Once we can hook up the Team API, we'll want to grab the start
+    # TODO:  date of an employee from there instead of defaulting to the
+    # TODO:  current date and time.
+
+    # TODO:  We'll also want to verify that the Slack user is an actual person
+    # TODO:  Against the Team API - Slack alone cannot tell us if they are real
+    # TODO:  or not.
+
+    employee = Employee.new(
```
jessieay: why not just `Employee.create` rather than `new` then `save`

app/services/employee_importer.rb
```diff
@@ -0,0 +1,74 @@
+require "slack-ruby-client"
+
+class EmployeeImporter
+  def initialize
+    configure_slack
+  end
+
+  def import(dry_run = false)
```
jessieay: thoughts on naming this `run` ? we are doing that (somewhat inconsistently) in other service classes

app/services/employee_updater.rb
```diff
@@ -0,0 +1,7 @@
+class EmployeeUpdater
+  def update
```
jessieay: thoughts on naming this `run` ?

jessieay:
OK @mrgilman I left a round of comments - let me know what you think!

app/services/employee_importer.rb
mrgilman: I was just following a pattern I saw elsewhere in the application; is there any reason not to put this in an initializer?

app/services/employee_importer.rb
mrgilman: The hash of import results is dependent on this returning `true` or `false`.

app/services/employee_importer.rb
mrgilman: Rubocop is enforcing this! Fixed now.

mrgilman:
@jessieay Thanks for the review! I've made most of the changes you suggested and replied to a few of your comments. Let me know what you think re: backfilling employees' `slack_user_id`s or skipping the update for employees where that field is blank.

app/services/employee_importer.rb
jessieay: no reason not to put in an initializer, in my opinion!

app/services/employee_importer.rb
jessieay: argh! I swear I took the thoughtbot guide and just copy pasted it....must have messed up. so sorry

mrgilman: Oh! I mean RuboCop was set up properly and I fixed the violation in this PR.

jessieay:
@mrgilman thank you! This looks good to me. Thank you for pointing out the issue with `employees` who have missing `slack_user_ids` - def makes sense to update those once daily for backfill.

@harrisj can you give this a review? Looks 👍  to me. I'd like to squash the commits before we merge.

lib/tasks/employees.rake
```diff
@@ -0,0 +1,41 @@
+namespace :employees do
+  desc "Imports all employees from Slack into the Dolores Landingham bot."
+  task import: :environment do
+    logger = ActiveSupport::Logger.new("log/employees_import.log")
+    start_time = Time.now
```
harrisj: This is also a good use for the `Benchmark.realtime` method, although that would require defining the `import_statistics` variable outside the block first, so I guess it's more of a tossup

mrgilman:
@jessieay Let me know if you'd like me to squash these commits. I'll probably leave @ccostino's as-is and squash mine together, if that sounds ok to you.

jessieay:
@mrgilman sure, sounds like a good squashing plan!

jessieay:
@harrisj this is squashed and ready for you to merge when you are 👍

harrisj:
Going to ignore the CodeClimate/Rubocop issues. This is good to me. Merging!

jessieay:
Tested the importer locally and it works like a charm!

ccostino:
I'm sorry I haven't been responsive in here, I only just had the chance to catch up, but thank you @mrgilman for jumping in and taking this on! :-)

Issue #125, opened 2016-01-06
Ability to send message to all Dolores users *right now*


melodykramer:
This is a slightly different use case for Dolores but would immensely help the @18F/comms and @18F/blog teams: the ability to message all Dolores users with the same message at once.

Use cases:

1. Alerting people to new features in Dolores
2. Asking people for information for the all-hands
3. Querying for upcoming blog posts or newsletters

The Comms team needs a way to communicate with each member of 18F on a regular basis. We've been individually DMing people, but that takes time and is not the easiest as 18F scales.

jessieay:
@melodykramer this would be pretty easy to implement.

melodykramer:
I think it would be a huge :+1:  for the @18F/blog team.

melodykramer:
(But I think the admin level should be there too. I'm not sure we'd want everyone to have access to this feature.)

awfrancisco:
So then all the replies go back to Dolores and the comms admin logs in as Dolores and manages all the replies individually?

jessieay:
We aren't currently logging responses sent to Dolores. Does this feature require that? Was thinking it was more of a one way thing. If we need to log responses, that's good to know.

melodykramer:
Yes, we would need a way to log the responses or collect them in some way.

gboone:
I'm not really sure how the sending of messages would work but maybe a "reply-to" feature where replies get sent to a group DM or a channel set by whomever triggers the sending.

jessieay:
@melodykramer if you could get in touch with blog / comms teams and figure out what the most helpful way to display responses would be (group DM or channel) that would be great. 

One limitation here is that I don't know if it will be possible send **certain** Mrs. Landingham replies to certain channels but not others. aka - we could have *all* replies forwarded to an email address, but we can't just have replies to a message sent from comms sent to an email address. Does that make sense? 

Another idea I had for dealing with replies would be to use webhooks, but those are currently only available in channels https://api.slack.com/outgoing-webhooks (not DMs)

So maybe a better way of dealing with this would be to say in the message body something like "DM Melody Kramer if you have any questions"

awfrancisco:
I think a channel would be better than a DM in case the Dolores monitoring team changes or the types of messages sent out change. Not sure how many unsolicited responses Dolores gets at the moment, but might be good to have a channel to drop those into anyway.

melodykramer:
Agree with @awfrancisco

melodykramer:
[...]

PR #193, opened 2016-07-19
[Connects #125] Send messages to all users

mrgilman:
I reused the existing `TestMessagesController` and `MessageSender` classes, which required adding some conditionals, but reduces duplication.


mrgilman:
Should the `last_sent_at` be displayed in EST/EDT? Sent messages are displayed in the time zone of the recipient; that does not apply here.

app/controllers/test_messages_controller.rb
```diff
@@ -1,27 +1,39 @@
...
+      @url = message_test_messages_path(@message)
+    end
   end
 
   def create
-    message = ScheduledMessage.find(params[:scheduled_message_id])
     employee = Employee.find_by(slack_username: slack_username)
 
-    if !employee.slack_channel_id.nil? || !employee.slack_user_id.nil?
+    if employee && (employee.slack_channel_id || employee.slack_user_id)
```
jessieay: ah yes we have some definite improvement to do in this controller! (pre-your contributions)

app/controllers/messages_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  def create
+    @message = Message.new(message_params)
+
+    if @message.save
+      flash[:notice] = "Message created successfully"
+      redirect_to messages_path
+    else
+      flash.now[:error] = "Could not create message"
```
jessieay: maybe include the actual message errors?

app/controllers/messages_controller.rb
```diff
@@ -0,0 +1,27 @@
+class MessagesController < ApplicationController
```
jessieay: can we add some `before_action`s in here to ensure that only an admin can create a message?

app/controllers/send_messages_controller.rb
```diff
@@ -0,0 +1,9 @@
+class SendMessagesController < ApplicationController
+  def create
```
jessieay: can we add a `before_action` to ensure that only admins can send these?

app/controllers/test_messages_controller.rb
```diff
@@ -1,27 +1,39 @@
...
+    else
+      flash[:error] = "Oops! Looks like this employees information isn't up to date. Check to make sure they haven't changed their slackname!"
     end
+
+    redirect_to redirect_path
   end
 
   private
 
+  def message
```
jessieay: is this method being used?

spec/features/send_test_message_spec.rb
```diff
@@ -14,8 +14,56 @@
...
+    )
+
+    login_with_oauth
+    visit messages_path
+
+    page.find(".button-test").click
+    fill_in "Slack username", with: username_from_fixture
+    click_on "Send test"
+
+    expect(page).to have_content("isn't up to date")
```
jessieay: note to self: we should put flash messages in I18n so they are easier to update and check for in tests

spec/features/send_message_spec.rb
```diff
@@ -0,0 +1,32 @@
+require "rails_helper"
+
+feature "Send message" do
+  scenario "message sends successfully", js: true do
```
jessieay: can we write a test that ensures a non-admin can't send a message to all employees? Want to make 100% sure non-admins can't spam the whole org ;)

app/controllers/test_messages_controller.rb
mrgilman: Yup, line 16.

app/controllers/send_messages_controller.rb
mrgilman: Ah, yup!

spec/features/send_message_spec.rb
mrgilman: Sure thing. I'll add the `before_action` and a controller spec.

app/controllers/test_messages_controller.rb
jessieay: ok cool - should we also use it for `def new` ?

mrgilman: Sure! Unfortunately we'll still need the conditional.

app/controllers/messages_controller.rb
mrgilman: Sure. The only validation is presence, and users won't be able to submit the form with blank fields, but I will add the error messages too just in case.

app/controllers/messages_controller.rb
mrgilman: Yup, done

mrgilman:
Hi @jessieay, thanks for the thorough review! I think I've addressed all of your feedback, and all of the Code Climate errors other than the frozen strings. I've been trying to fix a [flaky Timecop test](https://github.com/18F/dolores-landingham-bot/pull/193/files#diff-7fd990b0fb4fafe3b912dff8e1a04c11R14) but I think the current implementation should fix it.

I'm heading out of the office in a few minutes, and am working on a different project tomorrow, but let me know if there is anything else I can do to finish this up, I'm sure I can find some time.

jessieay:
This all lgtm! Passing the baton to @harrisj

mrgilman:
I'll squash so this is ready to go on 👍 .

jessieay:
Just tested this locally and received the message 4 times...let me see if this is something on my end or an issue where the code needs to be updated

jessieay:
One error I am seeing locally:

The import script imports *all* users, even bot users, which is fine by me but when we try to send a message to all users I see this error:

> "cannot_dm_bot

thoughts on how we can skip DMing bot users when we send these?

jessieay:
I've confirmed that I am receiving the messages multiple times. 

I think this is because some of the jobs are failing due to the above bot dm issue. The job runs for *all* users and succeeds for me but fails when it tries to DM the bot. So it retries and DMs me again.

Perhaps in addition to fixing that we should send each message in its own DJ so people don't receive multiple messages if the job fails when sending to a different user.

jessieay:
ps - if you rebase now, those code climate errors should be fixed ;)

jessieay:
@mrgilman hello! don't want to pressure you, but any update on fixing the "dm to bot" issue?

mrgilman:
Hi @jessieay! Planning to get back to finishing this up in the next hour. Sorry it didn;t work perfectly the first time around!

jessieay:
@mrgilman no problem! that's why we do manual QA :) thanks for the update

mrgilman:
@jessieay Is there any reason to store the bot users in the DB? If not, I'm leaning towards skipping the import, rather than storing and flagging them.

jessieay:
@mrgilman can't think of a reason to store them.

mrgilman:
@jessieay I made the suggested changes, and also removed an unused class and spec that I left in my other PR. Let me know how this works!

Also, awesome to see a green Code Climate build! 💚

jessieay:
@mrgilman woo! looks great to me. will let @harrisj take a final look tonight. thanks so much for the updates.

harrisj:
This looks good to me. I think in the future we might want to change names since it's a bit confusing at first to understand the difference between `Message` and `ScheduledMessage` (maybe the first could called `BroadcastMessage`). But this isn't really a demerit against the PR. In addition, it looks like we need to use some i18n in Dolores, but also not in scope of this auction. Good work!

jessieay:
ooooh @harrisj I like that naming suggestion! Going to create an issue for it. `Message` was my suggestion but you're right, it's not very clear.

PR #195, opened 2016-07-19
Do not check for frozen string literal comment in files

jessieay:
* Necessary for Ruby 3 but annoying right now
* https://github.com/bbatsov/rubocop/pull/2542


harrisj:
Boo, Rubocop!

Issue #181, opened 2016-05-27
Move to GitHub oauth


jessieay:
Myusa is on its way to being deprecated and is unfamiliar for new developers to this project to set up.

GitHub auth can be scoped to by team, so we can add GitHub oauth and only allow members of the 18F team to auth.

Added benefit: right now, anyone with an `@gsa` email address can auth on production, which is more people than we want.

jessieay:
Started work on this here https://github.com/18F/dolores-landingham-bot/compare/jy-github-auth?expand=1

jessieay:
PR here: https://github.com/18F/dolores-landingham-bot/pull/199

PR #199, opened 2016-07-21
Move to GitHub auth

jessieay:
* Using github team member auth: https://github.com/jonmagic/omniauth-github-team-member
* Only members of 18F team can log in
* Closes https://github.com/18F/dolores-landingham-bot/issues/181


CONTRIBUTING.md
```diff
@@ -112,58 +129,69 @@ you can contact Jessie Young. Otherwise, you can create a Slack bot
...
-  if is_permitted?(auth_email)
-
-  # Method
-  def is_permitted?(auth_email)
-    /#{ENV['AUTH_DOMAIN']}/.match(auth_email)
+  # app/controllers/auth_controller
+
+  def oauth_callback
+    # if team_member?
+      dev_test_email = your_email@example.com
```
yozlet: I realise that most people are going to know the right thing to do here, but how about putting that email address in quotes?

db/schema.rb
```diff
@@ -1,4 +1,3 @@
-# encoding: UTF-8
```
yozlet: What caused these `schema.rb` changes?

yozlet:
Added a couple of minor comments but neither of these is a blocker, just something to look at before you :shipit:. In other words, 👍

CONTRIBUTING.md
jessieay: hah goooood idea

db/schema.rb
jessieay: I think I forgot to run migrations after updating to Rails 5 earlier this week :(

Issue #198, opened 2016-07-21
Create "log in" page


jessieay:
Right now, authentication is required to view any and all pages. So if login fails, we don't have any pages where we can show the user the generated auth errors. 

We should create a simple page with a "login with github" that we can redirect users to if login fails. 

This can also serve as the non-authed root path.

jessieay:
Closed via https://github.com/18F/dolores-landingham-bot/pull/202 -- thanjs @timbo815 !

Issue #197, opened 2016-07-21
Add ability to log out


jessieay:
RIght not, only way to log out is to de-authorize to delete a cookie! Logout button in nav would be great.

jessieay:
Closed via https://github.com/18F/dolores-landingham-bot/pull/202 🌈

PR #202, opened 2016-07-23
added login button and ability to sign out

timbo815:
Fixes issue(s) #197 #198

-visiting "/" redirects to sign in page
-when signed in, sign out button is visible in sidebar
screen shot 2016-07-23 at 6 10 23 pm
screen shot 2016-07-23 at 6 10 34 pm

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -21,7 +21,7 @@
     </div>
 
     <ul class="usa-accordion sidenav-list">
-      <% if @current_user.admin? %>
+      <% if @current_user && @current_user.admin? %>
         <li class="sidenav-list-item">
           <button class="sidenav-list_title" aria-expanded="<%= is_current(users_path) -%>" aria-controls="collapsible-0">
             Users
@@ -29,7 +29,7 @@
             <icon class="icon-circle-minus"></icon>
           </button>
           <ul id="collapsible-0" aria-hidden="<%= !is_current(users_path) -%>" class="sidenav-sub_list usa-accordion-content">
-            <% if @current_user.admin? %>
+            <% if @current_user && @current_user.admin? %>
               <%= nav_link "View/Edit", users_path %>
             <% end %>
 
@@ -43,7 +43,7 @@
           <icon class="icon-circle-minus"></icon>
         </button>
         <ul id="collapsible-1" aria-hidden="<%= !is_current(scheduled_messages_path) -%>" class="sidenav-sub_list usa-accordion-content">
-          <% if @current_user.admin? %>
+          <% if @current_user && @current_user.admin? %>
             <%= nav_link "Create", new_scheduled_message_path %>
           <% end %>
           <%= nav_link "List/Edit", scheduled_messages_path  %>
@@ -56,7 +56,7 @@
           <icon class="icon-circle-minus"></icon>
         </button>
         <ul id="collapsible-2" aria-hidden="<%= !is_current(messages_path) -%>" class="sidenav-sub_list usa-accordion-content">
-          <% if @current_user.admin? %>
+          <% if @current_user && @current_user.admin? %>
             <%= nav_link "Create", new_message_path %>
           <% end %>
           <%= nav_link "List/Edit", messages_path  %>
@@ -84,7 +84,9 @@
         </ul>
       </li>
     </ul>
-
+    <% if current_user %>
+    <%= link_to "Sign Out", session_url, method: "delete" %>
+    <% end %>
 
   </aside>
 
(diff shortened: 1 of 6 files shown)


timbo815:
This failed some tests because I changed the root page to a dedicated sign in page. I am working on fixing that, and also writing some tests for the new approach. Let me know if you like the approach of having a dedicated sessions controller.

app/views/layouts/application.html.erb
```diff
@@ -21,15 +21,15 @@
     </div>
 
     <ul class="usa-accordion sidenav-list">
-      <% if @current_user.admin? %>
+      <% if @current_user && @current_user.admin? %>
```
jessieay: maybe we can create a method called `signed_in_as_admin?` that does both these checks

jessieay:
This looks like a great start! One thing I'd change regarding routing is have a different root path for signed in vs not signed in users. You can accomplish this with a constraint: https://www.viget.com/articles/using-routing-constraints-to-root-your-app

I like the idea of continuing to have a root path of `scheduled_messages#index` for logged in users and then using this new `sessions#new` root for nonlogged in users

jessieay:
FYI -- I am about to merge https://github.com/18F/dolores-landingham-bot/pull/199, which moves auth to GitHub. This shouldn't affect your code too much but it will affect your ability to test login and logout locally because you'll need to create a GitHub app in order to log in. I also added instructions on how to fake the auth locally if you'd like to do that. Just let me know if you have any questions!

app/views/layouts/application.html.erb
timbo815: agreed, much cleaner

timbo815:
Used a routing constraint as suggested to have separate root paths for signed in versus not signed in users

config/routes.rb
```diff
@@ -1,4 +1,14 @@
+class AuthConstraint
+  def matches?(request)
+    request.session[:user]["email"].present?
```
jessieay: maybe we should make sure that the email in the session matches a user in the db?

config/routes.rb
```diff
@@ -1,8 +1,18 @@
...
+  end
+end
+
 Rails.application.routes.draw do
-  root to: "scheduled_messages#index"
+  constraints(AuthConstraint.new) do
+    root to: "scheduled_messages#index"
```
jessieay: nice! can we write a little spec for this? maybe something similar to https://github.com/18F/dolores-landingham-bot/blob/master/spec/requests/auth_spec.rb but that makes sure a logged in user is taken to this page and a non-logged in user is not?

jessieay:
this looks super close! the code climate issue is just a whitespace issue,should be easy to clean up

config/routes.rb
timbo815: Sounds good, I'll get to work on that.

config/routes.rb
timbo815: That would work. Another thought I had was that we could add a `session_token` column to the users table. When a user signs in we can set their `session_token` using `SecureRandom.urlsafe_base64(16)`. We could then update the `current_user` method like so: `	@current_user ||= User.find_by(session_token: session[:session_token])` . On sign out the session token would be reset to nil. This might be more secure than using an email, let me know!

jessieay: one cool thing I learned recently is that rails automatically encrypts session tokens for you

> To prevent session hash tampering, a digest is calculated from the session with a server-side secret (secrets.secret_token) and inserted into the end of the cookie.

http://guides.rubyonrails.org/security.html#session-storage

so I think storing email is fine -- when we look it up it also automatically de-crypts so looking up user by email should be fine.

btw- learned this while having a very similar convo about `current_user` and security in C2 
https://github.com/18F/C2/blob/master/app/controllers/application_controller.rb#L118

timbo815: Oh awesome I didn't know that! Ok I will just do a check against the database with the email.

timbo815:
Let me know if you think this needs any improvements, thanks!

jessieay:
Just tested locally - looks awesome!!

I think for a next iteration I might use a different layout for `sessions#new` so that left sidebar doesn't show, but since we don't actually expose any data there, it works for now! Thanks so much. I am also excited about being able to sign out. ;) 💯 

Jessie

jessieay:
pushed to staging and prod and works like a charm -- woo hoo!

timbo815:
Awesome! Do you want me to make a nice sign in page that doesn't display the sidebar and with CSS styling?

jessieay:
@timbo815 if you'd like to that would be awesome! No pressure, though.

PR #208, opened 2016-08-28
just enough to get vagrant up to work

eyefodder:
Fixes issue(s) # .

- added to .gitignore to ignore .vagrant folder created on ```vagrant up```
- added ```git``` as part of vagrant provisioning script
- fixed typo ```cp .sample.env .env```

/cc relevant people


afeld:
Confirmed this works by pairing with @eyefodder. Will merge now to fix the problems for other people getting set up, but we can certainly do any requested changes as a follow-up. Thanks for contributing!

PR #214, opened 2016-09-15
Fix issue where east coast users didn't get messages after 4:00 PM

### LINKED ISSUE #111: Messages cannot be sent out late at night
It appears there is another issue related to the employee message matching when sending messages late at night, this time in checking the start date of the employee.  After a certain point in time in the evening (I was testing between 9:00 PM and 10:00 PM) two tests start to fail:

* MessageEmployeeMatcher#run matches messages to users based on days after start
* MessageEmployeeMatcher#run ignores messages that have already been sent to users

And messages will not be sent when testing things manually.

I suspect this is a similar issue that @jessieay and I had fixed with the time matching where we also had to check that the user's local time was still during the same day as the UTC time the app uses internally.  Checking to see which day to send a message out may require a similar guard clause in it.

jmhooper:
closes #111

Proposed changes

- Adjust `retrieve_employees_needing_onboarding_message` to query for employees that could potentially receive messages on the given day outside of the UTC timezone by spanning the next and previous days
- Refactor `time_to_send_message?` to calculate the time a user should receive a message and compare that to the current time instead of comparing days/hours/minutes separately
- Fix an erroneous spec so that unsent messages that were scheduled to be sent in the previous evening can be sent after their scheduled time

### DIFF
--- a/app/services/message_employee_matcher.rb
+++ b/app/services/message_employee_matcher.rb
@@ -22,8 +22,13 @@ class MessageEmployeeMatcher
   end
 
   def retrieve_employees_needing_onboarding_message
-    Employee.where(started_on: day_count.business_days.ago).select do |employee|
-      time_to_send_message?(employee.time_zone) &&
+    Employee.where(
+      started_on: Range.new(
+        day_count.business_days.ago - 1.day,
+        day_count.business_days.ago + 1.day,
+      )
+    ).select do |employee|
+      time_to_send_message?(employee) &&
         onboarding_message_not_already_sent?(employee)
     end
   end
@@ -36,17 +41,14 @@ class MessageEmployeeMatcher
     message.days_after_start
   end
 
-  def time_to_send_message?(time_zone)
-    employee_current_time = Time.current.in_time_zone(time_zone)
-
-    if employee_current_time.day == Time.current.day
-      employee_current_time_value = employee_current_time.strftime("%H%M").to_i
-      message_time_value = message.time_of_day.strftime("%H%M").to_i
-
-      employee_current_time_value >= message_time_value
-    else
-      false
-    end
+  def time_to_send_message?(employee)
+    send_date_without_timezone = day_count.business_days.after(employee.started_on)
+    send_date_with_timezone = send_date_without_timezone.in_time_zone(employee.time_zone)
+    send_time = send_date_with_timezone.change(
+      hour: message.time_of_day.hour,
+      min: message.time_of_day.min,
+    )
+    Time.current >= send_time
   end
 
   def onboarding_message_not_already_sent?(employee)
--- a/spec/services/message_employee_matcher_spec.rb
+++ b/spec/services/message_employee_matcher_spec.rb
@@ -30,7 +30,7 @@ describe MessageEmployeeMatcher do
           Timecop.freeze(Time.parse("00:00:00 UTC")) do
             days_after_start = 3
             scheduled_message_time = Time.parse("12:00:00 UTC")
-            scheduled_message = create(:scheduled_message, days_after_start: days_after_start, time_of_day: scheduled_message_time)
+            scheduled_message = create(:scheduled_message, days_after_start: days_after_start + 1, time_of_day: scheduled_message_time)
 
             _employee_cst = create(:employee, started_on: days_after_start.business_days.ago, time_zone: "Central Time (US & Canada)")
 


spec/services/message_employee_matcher_spec.rb
```diff
@@ -5,7 +5,7 @@
   describe "#run" do
     context "onboarding messages (sent days after start)" do
       it "matches messages to users based on days after start" do
-        Timecop.freeze(Time.parse("11:00:00 UTC")) do
+        Timecop.freeze(Time.parse("0:00:00 UTC")) do
```
jessieay: we have a context below that is context "midnight UTC" do` -- should we combine these specs into that context? 

trying to think about how we can make it obvious that our tests cover this case so regressions aren't added in the future.

spec/services/message_employee_matcher_spec.rb
```diff
@@ -30,7 +30,7 @@
           Timecop.freeze(Time.parse("00:00:00 UTC")) do
             days_after_start = 3
             scheduled_message_time = Time.parse("12:00:00 UTC")
-            scheduled_message = create(:scheduled_message, days_after_start: days_after_start, time_of_day: scheduled_message_time)
+            scheduled_message = create(:scheduled_message, days_after_start: days_after_start + 1, time_of_day: scheduled_message_time)
```
jessieay: what does this change mean? I think it might no longer match the spec description with this change.

the test description is ""does not send messages to employees who are at a later time but on a different date" 

In this changed spec, do we still have a user who is at a later time but on a different date?

I can already feel my brain getting fried just thinking about this. This may be correct, just want to make sure I understand the implications.

app/services/message_employee_matcher.rb
```diff
@@ -22,8 +22,13 @@ def retrieve_matching_employees
   end
 
   def retrieve_employees_needing_onboarding_message
-    Employee.where(started_on: day_count.business_days.ago).select do |employee|
-      time_to_send_message?(employee.time_zone) &&
+    Employee.where(
+      started_on: Range.new(
```
jessieay: ooh range. Don't get to use those much.

app/services/message_employee_matcher.rb
```diff
@@ -36,17 +41,13 @@ def day_count
...
-    if employee_current_time.day == Time.current.day
-      employee_current_time_value = employee_current_time.strftime("%H%M").to_i
-      message_time_value = message.time_of_day.strftime("%H%M").to_i
-
-      employee_current_time_value >= message_time_value
-    else
-      false
-    end
+  def time_to_send_message?(employee)
+    employee_start_date = day_count.business_days.after(employee.started_on)
```
jessieay: I am not sure what we have here is an employee start date (unless I am misunderstanding the code here, which is totally possible)...is there a better var name we can use than ` employee_start_date`

app/services/message_employee_matcher.rb
```diff
@@ -36,17 +41,13 @@ def day_count
...
+    send_time = employee_start_date.in_time_zone(employee.time_zone).change(
+      hour: message.time_of_day.hour,
+      min: message.time_of_day.min,
+    )
+    Time.now >= send_time
```
jessieay: I think `Time.current` is in UTC. `now` not necessarily http://stackoverflow.com/a/12748804/1019369 (probably want `current`?)

spec/services/message_employee_matcher_spec.rb
jmhooper: This is a debug artifact. I'll change it back. Just did locally and the specs are still green.

spec/services/message_employee_matcher_spec.rb
jmhooper: My brain is also fried 🍳

My understanding is that what the spec is protecting against is users who should receive a message at `2016/1/2 17:00 PT` receiving the message at `2016/1/1 4:00:00 PST` since that is midnight of the next day in UTC and method was previously only comparing minutes and hours.

Prior to this commit, the spec would create a message to go out at `2016/1/1 12:00:00 PST` and confirm that it was not sent at `2016/1/2 00:00:00 UTC`. The message should, however, be sent in that situation (if it has not already) because `2016/1/2 00:00:00 UTC > 2016/1/1 12:00:00 PST`.

app/services/message_employee_matcher.rb
jmhooper: Oops. You're right.

Maybe `employee_message_date` since it is the date the message should be sent, agnostic of any TZ or time info? Tricky name...

jessieay: I feel like `employee_message_date` still feels pretty vague. how about `employee_day_count_in_employee_time_zone` -- is that an accurate way to represent this value?

app/services/message_employee_matcher.rb
jmhooper: I don't think it will matter, but in my experience `Time.current` is preferred over `Time.now` so I'll switch it.

app/services/message_employee_matcher.rb
jmhooper: Hmm. `day_count` makes me this should be an integer though since the service has another private method named `day_count` which returns an integer.

Maybe `employee_date_in_employee_time_zone`?

jmhooper: @jessieay: Okay, I thought about this some with a fresh brain and here's what I came up with:

```ruby
    send_date_without_timezone = day_count.business_days.after(employee.started_on)
    send_time_with_timezone = send_date_without_timezone.in_time_zone(employee.time_zone).change(
      hour: message.time_of_day.hour,
      min: message.time_of_day.min,
    )
    Time.current >= send_time_with_timezone
```

Changed it to that in my latest fixup

jmhooper: @jessieay: Broke it out into something that is a bit more readable. The second line is still > 80, but not by as much:

```ruby
send_date_without_timezone = day_count.business_days.after(employee.started_on)
send_date_with_timezone = send_date_without_timezone.in_time_zone(employee.time_zone)
send_time = send_date_with_timezone.change(
  hour: message.time_of_day.hour,
  min: message.time_of_day.min,
)
Time.current >= send_time
```

spec/services/message_employee_matcher_spec.rb
jessieay: This sounds RIGHT! 🌈

jessieay:
Looks ready to squash n merge!

PR #215, opened 2016-09-15
Add "Forbidden" page for unsuccessful auth

jmhooper:
Fixes issue(s) # 213

- Add error page that is displayed after unsuccessful sign in

This commit adds a page that is displayed when the user attempts to authenticate and fails.

The page provides some hints about where to look to determine what could be causing the problem.
Additionally it provides a link to attempt another sign in.

Originally I was going to render the new session view with a flash message, but decided on a separate page since it provides more information about what might have gone wrong.

### DIFF
--- a/app/views/auth/forbidden.html.erb
+++ b/app/views/auth/forbidden.html.erb
@@ -1,0 +1,17 @@
+<section class="main-content">
+  <h1>Forbidden</h1>
+  <p>
+    We were unable to authenticate your user profile.
+  </p>
+  <p>
+    If you are signing in via GitHub, please verify that you are a member of the
+    appropriate GitHub organization and that your
+    <%=
+      link_to "membership is public",
+        "https://help.github.com/articles/publicizing-or-hiding-organization-membership/"
+    %>.
+  <p/>
+  <p>
+    <%= link_to "Try to sign in with GitHub again", "/auth/githubteammember" %>
+  </p>
+</section>
--- a/spec/support/oauth_helper.rb
+++ b/spec/support/oauth_helper.rb
@@ -4,11 +4,11 @@ module OauthHelper
     visit "/auth/githubteammember"
   end
 
-  def setup_mock_auth(email = "test@example.com")
+  def setup_mock_auth(email = "test@example.com", options = {})
     OmniAuth.config.mock_auth[:githubteammember] =
       OmniAuth::AuthHash.new(
         credentials: {
-          "team_member?" => true,
+          "team_member?" => team_member?(options),
         },
         provider: "githubteammember",
         info: {
@@ -18,5 +18,15 @@ module OauthHelper
           uid: "12345",
         },
     )
+  end
+
+  private
+
+  def team_member?(options)
+    if options[:team_member].nil?
+      true
+    else
+      options[:team_member]
+    end
   end
 end
--- a/app/controllers/auth_controller.rb
+++ b/app/controllers/auth_controller.rb
@@ -7,6 +7,8 @@ class AuthController < ApplicationController
       sign_in(user)
       flash[:success] = I18n.t('controllers.auth_controller.successes.oauth_callback')
       redirect_to root_path
+    else
+      render :forbidden
     end
   end
 
--- a/spec/features/sign_in_user_spec.rb
+++ b/spec/features/sign_in_user_spec.rb
@@ -1,0 +1,41 @@
+require "rails_helper"
+
+feature "Sign in user" do
+  scenario "signing in a new user that is a member of the correct Github team" do
+    setup_mock_auth("user@example.com")
+
+    visit root_path
+    click_link "Sign in with GitHub"
+
+    expect(page).to have_content "You successfully signed in"
+  end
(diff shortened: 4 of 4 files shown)


jessieay:
Nice! Can we add a test for this?

jmhooper:
Definitely. Should this be covered by a controller spec, request spec, or both?

jessieay:
I'd be happy with either *or* a capybara feature spec

jmhooper:
@jessieay: Added some tests ✅✅

Weirdly enough it looks like one of the CI builds failed b/c of a spec that I didn't touch on this branch, but did on my other branch.

app/views/auth/forbidden.html.erb
```diff
@@ -0,0 +1,13 @@
...
+  <h1>Forbidden</h1>
+  <p>
+    We were unable to authenticate your user profile.
+  </p>
+  <p>
+    If you are signing in via Github, please verify that you are a member of the appropriate Github organization and that your
+    <%= link_to "membership is public", "https://help.github.com/articles/publicizing-or-hiding-organization-membership/" %>.
+  <p/>
+  <p>
+    <%= link_to "Try to sign in with Github again", "/auth/githubteammember" %>
```
jessieay: `GitHub` :)

spec/support/oauth_helper.rb
```diff
@@ -4,9 +4,9 @@ def login_with_oauth(user = create(:user))
     visit "/auth/githubteammember"
   end
 
-  def setup_mock_auth(email = "test@example.com")
+  def setup_mock_auth(email = "test@example.com", overrides = {})
```
jessieay: do we need overrides?

app/views/auth/forbidden.html.erb
```diff
@@ -0,0 +1,13 @@
+<section class="main-content">
+  <h1>Forbidden</h1>
+  <p>
+    We were unable to authenticate your user profile.
+  </p>
+  <p>
+    If you are signing in via Github, please verify that you are a member of the appropriate Github organization and that your
```
jessieay: `GitHub`

jessieay: can we break these lines at 80 chars por favor?

spec/features/create_user_spec.rb
```diff
@@ -0,0 +1,27 @@
+require "rails_helper"
+
+feature "Create user" do
```
jessieay: maybe we should make these more feature-y. So "User signs in" instead of "Create user" and "user is member of the correct GitHub team" rather than "creating a user..." (also perhaps we should rename file)

spec/features/create_user_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+    setup_mock_auth("user@example.com", {
+      credentials: {
+        team_member?: false,
+      },
+    })
+
+    visit root_path
+    click_link "Sign in with GitHub"
+
+    print page.html
```
jessieay: debugger!

jessieay:
Oh yeah,I think that was another issue we discovered with code: CI fails in evening 😆 
Glad you opened that FIx!

jmhooper:
@jessieay: I changed `create_user_spec.rb` to `sign_in_user_spec.rb` and added some specs to cover signing in existing users.

I also removed the debug artifact.

spec/support/oauth_helper.rb
jmhooper: I added them to test unsuccessful auth, but maybe I could change it to an `options` argument that handles something like `team_member?: false`?

jmhooper:
Okay, just pushed another fixup

- addressed the comment about the `overrides` hash by making it into an `options` hash
- addressed some of the issues in the auth/forbidden view

Also, TIL I've been spelling "GitHub" as "Github" my entire life without realizing it's supposed to be CamelCase 😳.

jessieay:
@jmhooper LOL CamelCase, sooooo oldschool GitHub

spec/support/oauth_helper.rb
```diff
@@ -4,11 +4,11 @@ def login_with_oauth(user = create(:user))
     visit "/auth/githubteammember"
   end
 
-  def setup_mock_auth(email = "test@example.com")
+  def setup_mock_auth(email = "test@example.com", options = {})
     OmniAuth.config.mock_auth[:githubteammember] =
       OmniAuth::AuthHash.new(
         credentials: {
-          "team_member?" => true,
+          "team_member?" => options[:team_member].nil? ? true : options[:team_member],
```
jessieay: we can change this in the future (no need to do it now), but I generally avoid these types of ternary statements. Reason: They are nice and compact but can be harder to understand.

jessieay:
Good one! this Looks great. Do you mind squashing and merging? That is not the process on every 18F project, but it's what we've been doing on this one: https://github.com/18F/dolores-landingham-slack-bot/blob/master/CONTRIBUTING.md#git-protocol

If you are not comfortable with squashing (can be scary) I am happy to walk you through it.

Issue #206, opened 2016-08-02
Consider breaking out `Quarterly message` from `ScheduledMessage`


jessieay:
right now, quarterly messages are a type of scheduled message. You create them in the same form, there is just a checkbox you check if you want something to be a quarterly message

It would be more obvious from an admin standpoint how to create a quarterly message if it had its own section in the lefthand nav.  This means a separate form (with no checkbox required, and only fields that you need for a quarterly message) and a separate list view for quarterly messages.

This might not mean we need to change anything at the admin level, but we could have views for each type that are separate. I am also open to separating them out in the database, though.

jessieay:
Closed via https://github.com/18F/dolores-landingham-slack-bot/pull/223 -- guess our template does not have language that auto-closes! jeez

PR #223, opened 2016-09-21
Break ScheduledMessage out into OnboardingMessage and QuarterlyMessage

jmhooper:
Fixes issue(s) #206

- Break `ScheduledMessages` model into `QuarterlyMessage` model and `OnboardingMessage` model
- Change `SentScheduledMessage` model to `SentMessage` and make the relationship to messages polymorphic

This PR is WIP


app/assets/javascripts/scheduled_message.js
```diff
@@ -1,15 +0,0 @@
-$(document).ready(function() {
```
jessieay: yess so happy to be done with this.

app/models/quarterly_message.rb
```diff
@@ -0,0 +1,30 @@
...
+  acts_as_taggable
+
+  has_many :sent_messages, as: :message, dependent: :destroy
+
+  validates :body, presence: true
+  validates :tag_list, presence: true
+  validates :title, presence: true
+
+  # disable STI
+  self.inheritance_column = nil
```
jessieay: do we need this now that there is no longer a `type` column?

db/chores/migrate_messages_to_scheduled_messages.rb
```diff
@@ -0,0 +1,91 @@
...
+      migrate_quarterly_message_to_scheduled_message(row)
+    end
+  end
+
+  private
+
+  def migrate_onboarding_message_to_scheduled_message(row)
+    # Migrate to new table
+    row["type"] = 0
+    values = ["created_at","updated_at","title","body","days_after_start","time_of_day","type","end_date","deleted_at"].map do |column|
```
jessieay: small style thing: would it be possible to put each array val on its own line? I think that would be easier to read

db/chores/migrate_messages_to_scheduled_messages.rb
```diff
@@ -0,0 +1,91 @@
...
+      sanitize(row[column])
+    end
+    scheduled_message_id = insert(
+      <<-SQL
+        INSERT INTO scheduled_messages (created_at, updated_at, title, body, days_after_start, time_of_day, type, end_date, deleted_at)
+        VALUES (#{values.join(",")})
+      SQL
+    )
+
+    # Update sent messages
```
jessieay: what about putting each of these into a well-named method rather than using comments? eg: lines 27 through 33 can be `def update_sent_messages` ?

db/chores/migrate_messages_to_scheduled_messages.rb
```diff
@@ -0,0 +1,91 @@
...
+        UPDATE taggings
+        SET taggable_id = #{quarterly_message_id}, taggable_type = 'ScheduledMessage'
+        WHERE taggable_id = #{row["id"]} AND taggable_type = 'QuarterlyMessage'
+      SQL
+    )
+  end
+
+  # Helpers
+
+  def execute(*args)
```
jessieay: not necessary but it might be good to put these into a shared place since you use them in two places 🌞

spec/models/onboarding_message_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+     it { should validate_presence_of(:body) }
+     it { should validate_presence_of(:days_after_start) }
+     it { should validate_presence_of(:tag_list) }
+     it { should validate_presence_of(:time_of_day) }
+     it { should validate_presence_of(:title) }
+  end
+
+  describe '.date_time_ordering' do
+    it 'should display the messages in chronological order' do
+      m1 = create(:onboarding_message, days_after_start: 1, time_of_day: '2000-01-01 11:00:00 UTC')
```
jessieay: can we create these out of order so we know that the ordering is working?  I think the would return in this order by default anyway so that would increase the quality of test

jmhooper: Yes, good idea

db/chores/migrate_messages_to_scheduled_messages.rb
jmhooper: I thought about that. Maybe something like a `ScheduledMessageSQLMethods` module that both of these could include?

db/chores/migrate_messages_to_scheduled_messages.rb
jmhooper: I can. The reason I did this was so that in dev I could line them up visually with the column names in the SQL a couple lines down.

app/models/quarterly_message.rb
jmhooper: We do not. I'll clean up.

app/assets/javascripts/scheduled_message.js
jmhooper: 👋

db/chores/migrate_messages_to_scheduled_messages.rb
jessieay: sure, sounds great

db/chores/migrate_messages_to_scheduled_messages.rb
jessieay: fine as-is if you prefer it!

jmhooper:
> I'd love to see some unit tests for the chores classes so we have confidence that the migrate data the way we want them to!

This was tripping me up the other night. The chores require the schema to be in a weird transitional state to work. For example, to migrate the scheduled messages to quarterly/onboarding messages, you need the `scheduled_messages` table, the `quarterly_messages` table, and the `onboarding_messages` table. The schema is not in that state before or after the migrations.

The migration tests I was looking at involved rolling back to a particular migration, adding some data, running the next migration, and testing the result. I'm not sure how to do something like that for the chores since necessary state for the schema isn't achievable by rolling back to a particular migration.

My thought was to roll back to the previous migration, grab the code from the `up` and `down` methods of the migration before the chores are called, and move it into a `before_each` hook in spec. I'm not a huge fan of that solution though b/c it seems like there should be a Better Way™

Any thoughts?

jessieay:
@jmhooper ah yes, that is super complicated. Did you try out the technique described in http://blog.carbonfive.com/2011/01/27/start-testing-your-migrations-right-now/ (which you linked me to)? 

I've never used something like that before, but looks interesting! It also looks like the strategies they describe involve running the migration. I think that if you went back to the previous migration version (`ActiveRecord::Migrator.migrate @previous_version` in that blog post), created some records with the old DB schema, then ran the migration and confirmed that the new records that you'd expect exist, that would be more than good enough w/r/t testing.

jmhooper:
@jessieay: Addressed some of your comments and pushed some more commits to round this out.

The status of this can change from WIP to Ready for Review.

app/models/onboarding_message.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  def self.active
+    where("end_date IS NULL OR end_date > ?", Date.today)
+  end
+
+  def self.date_time_ordering
+    order(:days_after_start, :time_of_day)
+  end
+
+  def self.filter(params)
```
jessieay: since this is exactly the same here and in `QuarterlyMessage` -- thoughts on putting it in a shared concern? http://api.rubyonrails.org/classes/ActiveSupport/Concern.html

app/models/sent_message.rb
```diff
@@ -0,0 +1,54 @@
...
+  validates :sent_at, presence: true
+  validates :sent_on, presence: true
+
+  delegate :slack_username, to: :employee
+
+  def self.by_year(year)
+    where("extract(year from created_at) = ?", year)
+  end
+
+  def self.filter(params)
```
jessieay: some comment here re: moving shared logic into a concern

app/views/onboarding_messages/edit.html.erb
```diff
@@ -0,0 +1,6 @@
+<%= render 'header', :title => 'Update onboarding message' %>
+
+<section class="main-content">
+	<%= render "flashes" -%>
```
jessieay: maybe flashes should be in layout? (this could be done via a separate issue/PR)

app/models/onboarding_message.rb
jmhooper: My thought was to open an issue once this was merged to move it into a query object. I can go ahead and do that or move it into a concern in this PR though.

app/views/onboarding_messages/edit.html.erb
jmhooper: I agree. I can open an issue.

app/models/onboarding_message.rb
jessieay: sure! good plan, open an issue.

PR #224, opened 2016-09-21
Fixes #216, adds partials for application layout and adjusts sidenav …

### LINKED ISSUE #216: Sidebar appears on Sign In pages
Currently the app has 2 views where the user is in an unauthenticated state:

- The Sign In page (`session/new`)
- The Sign In Error page (`auth/forbidden`)

Both of those views are rendered in the application layout which makes the assumption that the user is signed in and displays the sidebar.

image

image

We need to hide the sidebar when the user is not authenticated. A few options:

- render the unauthenticated pages in a different layout
- render them without a layout
- hide the sidebar in the application layout when the user is not signed in

I'm a fan of the first but invite discussion.

For more on this, reference the discussion on #213

JVThomas:
this address #216

- Added if statements in application.html.erb to adjust sidenav links based on auth status
- Broke up sidenav list items into two different partials, one for admin only items and another for items for general logged in status

Notes

- I made the partials because I thought the original application.html.erb file was a bit large and hard to navigate, so I broke it down to organize it better
- Sidenav item visibility changes depending on login status.

### DIFF
--- a/app/views/layouts/_admin_sidenav_items.html.erb
+++ b/app/views/layouts/_admin_sidenav_items.html.erb
@@ -1,0 +1,12 @@
+<li class="sidenav-list-item">
+  <button class="sidenav-list_title" aria-expanded="<%= is_current(users_path) -%>" aria-controls="collapsible-0">
+    Users
+    <icon class="icon-circle-plus"></icon>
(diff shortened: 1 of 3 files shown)


jessieay:
This looks great! thank you for submitting. 🌈

PR #232, opened 2016-09-27
Remove `--deployment` from travis's bundler options

jmhooper:

- Remove `--deployment` from travis's bundler options

Bundler's `--deployment` flag causes gems to be installed in `vendor/bundle`. This causes problems with automated deploys since some of these files have a mode < 600.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 before_script:
 - cp .sample.env .env
 - bin/rake db:setup --trace
-bundler_args: "--deployment --jobs=3 --retry=3 --without development"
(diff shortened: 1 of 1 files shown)


jmhooper:
You can read about [bundler's deployment mode here](http://bundler.io/v1.1/man/bundle-install.1.html#DEPLOYMENT-MODE). It is designed to isolate your app by preventing it from sharing gems with the system. As a result, it causes bundler to produce a bunch of files with file mode 444 which CloudFoundry does not allow.

jessieay:
Oh interesting, so why do we think it was working before?

PR #233, opened 2016-09-28
Add vendor/bundle to gitignore

jmhooper:

- Add vendor/bundle to ~~Gemfile~~ gitignore

This commit works to enable automated deploys from Travis.

Deploys were failing because of files with a file mode of 444 being present on deploy to Cloud Foundry.
Cloud Foundry does not allow a deploy with files that have a mode < 600.

After some investigation, it was discovered that the files in question were in `/vendor/bundle` since bundler was being run in deployment mode on CI.
Deployment mode was disabled in 9b5c13104c5e532c52c65bb619b4965f5a9169da, but the builds still failed to deploy.

Some more investigation yielded the following from Travis's docs:

> If you have custom Bundler arguments, and these include the –path option, Travis CI will use that path. If –path is missing but –deployment is present, it will use vendor/bundle.
>
> Otherwise it will automatically add the –path option. In this case it will either use the value of the environment variable BUNDLE_PATH or, if it is missing, vendor/bundle.

ref: https://docs.travis-ci.com/user/caching/#Bundler

So, regardless of whether bundler is in deployment mode or not, the dependencies are still written to `vendor/bundle`.
This commit attempts to fix this problem by adding `vendor/bundle` to the gitignore.

This rule is present in GitHub's default Rails gitignore: https://github.com/github/gitignore/blob/master/Rails.gitignore#L27


jessieay:
Hey! the title of this PR is "Add vendor/bundle to Gemfile" -- did you mean "Add vendor/bundle to gitignore" ?

Just want to make sure bc I am not sure I understand the title as-is

Also, I am usually not in favor of code comments but this seems like a case where we might forget this history of this line in the future. Perhaps add a comment with a link to this PR next to that line in `.gitignore` ?

jmhooper:
> Hey! the title of this PR is "Add vendor/bundle to Gemfile" -- did you mean "Add vendor/bundle to gitignore" ?

Ha, yeah I meant gitignore

> Also, I am usually not in favor of code comments but this seems like a case where we might forget this history of this line in the future. Perhaps add a comment with a link to this PR next to that line in .gitignore ?

Yes, I'll stick one in there.

jmhooper:
@jessieay: Okay, I amended the commit to fix the message and add a comment to the gitignore.

Issue #234, opened 2016-10-07
Make quarterly messages work for TTS Slack org


jessieay:
Last week, 18F's Slack org was moved from being for 18F only to being for all of TTS.

While our Slack now includes more people, our quarterly messages are sometimes meant to be for 18F team members only. 

Quarterly Messages are messages that will be Slack DM'd to every employee that is in the Dolores database on the first day of each quarter.

The original issue for this feature is here: https://github.com/18F/dolores-landingham-slack-bot/issues/180

We regularly import all Slack users in our Slack org into the Dolores database. As far as I know, there is currently no way to know if a Slack user is from 18F or not. So our nightly import will also include non-18F TTS employees who are in Slack.

Before the ops team creates any quarterly messages, we need to decide how to move forward.

Ideas for how to do this:
- use 18F's [team API](https://team-api.18f.gov/public/api/team/) to cross-check which Slack users are from 18F (not sure if the Team API is up to date)
- Manually mark employees in the Dolores DB as "18F employees" (or maybe mark people as NOT 18F employees, since right now we have fewer non-18F people than 18F people in our Slack org)
- Think of some other smart way to do this :)

Related question: should we disable the ability to even create a quarterly message until this is resolved?

jmhooper:
I saw some murmurings about creating an 18F channel for news/announcements specific to 18F. If that happens maybe we could import users who are subscribed to that channel?

joshuabailes:
I like the @jmhooper approach. Anyone know if that would work in Slack?

jessieay:
good idea @jmhooper -- want to follow up on these murmurings and see what you find? I have been lurking in `#wg-slack` but nothing about an 18F-only team yet...

PR #235, opened 2016-10-12
Temporarily disable quarterly messages

jmhooper:

- Comment out quarterly message code in clockwork config
- Add notice about quarterly messages being disabled to views

This temporarily disables quarterly messages to prevent confusion / miscommunication until we move forward on #234.

### DIFF
--- a/app/assets/stylesheets/components/_alert.scss
+++ b/app/assets/stylesheets/components/_alert.scss
@@ -4,7 +4,6 @@
   background-color: #d0e9c6;
   font-size: 0.8rem;
   padding: 0.5em;
-  @include span-columns(10);
   border-radius:0.5em;
   margin-bottom: 1em;
 }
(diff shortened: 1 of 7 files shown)


jmhooper:
Here's what the quarterly message views look like now:

image

app/assets/stylesheets/components/_alert.scss
```diff
@@ -4,7 +4,6 @@
   background-color: #d0e9c6;
   font-size: 0.8rem;
   padding: 0.5em;
-  @include span-columns(10);
```
jessieay: I think it is used automatically when we have a `flash` but could be wrong

jmhooper: I had the same thought, and when I looked it seemed like those are rendered with the class `flash-alert`:

```erb
<div class="flashes">
  <% user_facing_flashes.each do |key, value| -%>
    <div class="flash-<%= key %>"><%= value %></div>
  <% end -%>
</div>
```

Ref: https://github.com/18F/dolores-landingham-slack-bot/blob/c88f124da52a741db47c66e4e44d529eab36862c/app/views/application/_flashes.html.erb#L4

The styles for that seems to live in `app/assets/stylesheets/components/_flashes.scss`.

Ref: https://github.com/18F/dolores-landingham-slack-bot/blob/c88f124da52a741db47c66e4e44d529eab36862c/app/assets/stylesheets/components/_flashes.scss

jessieay:
LGTM! Thanks for doing this. Merging.

PR #249, opened 2017-03-08
Send delayed jobs

jessieay:
* Delayed jobs were not working before. Now they are!

cc @davomm 

Closes https://github.com/18F/dolores-landingham-slack-bot/issues/248


jessieay:
cc @jmhooper

jessieay:
@jmhooper good idea! I will have to ssh to confirm, but there may be none right now since I believe we only call delay for test messages and broadcast messages. Regular message are sent in a clock process so we don't send them as delayed jobs.

PR #254, opened 2017-03-10
Remove Employee uniqueness validation from SentMessage

jmhooper:
Fixes issue(s) # 250

- Remove Employee uniqueness validation from SentMessage

The SentMessage model had a uniqueness validation scoped to the message. The effect of this validation is that a SentMessage can only be created once for a given message and employee. As a result, sending a message to an employee twice resulted in a failure to create a SentMessage record for the second message.

This commit fixes the above bug by removing the validation.

A byproduct of this change is that if an employee has been sent an onboarding message as part of a test before they were supposed to receive that message, they will not receive it at the scheduled time. This does not seem like a major concern since the employee will have ultimately still received the message.

Another byproduct is that an employee that has been sent a test quarterly message within a week of the beginning of a new quarter will not receive the message at the start of the new quarter. This does not seems like a concern since the employee will ultimately still receive receive the message near the beginning of the new quarter.

Finally this fix makes the assumption that there is not a reason to avoid creating SentMessage records for test messages. I am not aware of one, but if one exists, a follow up commit may be needed to tell the MessageSender not to create a SentMessage record for test messages. I think the fix on this commit still applies in that case, since we will still want to keep a record of when non-test messages get sent, even if they are sent twice for some reason.

/cc @jessieay

### DIFF
--- a/app/models/sent_message.rb
+++ b/app/models/sent_message.rb
@@ -4,9 +4,7 @@ class SentMessage < ActiveRecord::Base
   belongs_to :employee
   belongs_to :message, polymorphic: true
 
-  validates :employee, presence: true, uniqueness: {
-    scope: [:message_id, :message_type],
-  }
+  validates :employee, presence: true
   validates :message_body, presence: true
   validates :message, presence: true
   validates :sent_at, presence: true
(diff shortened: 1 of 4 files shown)


jessieay:
HI! I think this is a good direction to go, thanks for the detailed commit message. 2 changes I think we need:

1. Migration to roll back the uniqueness index at the postgres level: https://github.com/18F/dolores-landingham-slack-bot/blob/754af081cb7f0d2d4ba22720ec859d515efd7fdf/db/schema.rb#L87

2. Feature spec for a regression test to make sure we don't run into this situation again. So in the spec, send a test message and then sure that the user can receive the same message as an onboarding message, too.

jmhooper:
Amended the commit. This should be ready to go

Issue #250, opened 2017-03-08
Issue with testing same message to same person


jessieay:
in `app/services/message_sender.rb`, we create a `SentMessage` record for each message sent. This is so that we can record who receives what and when. 

Right now, we validate the uniqueness of the employee id, message id, and message type for each sent message. So, if we try to test the same message to the same employee, that `SentMessage` creation while fail on the second try. 

We do not need to create `SentMessage` records in the testing context. We need to find a way around this.

jessieay:
This issue will also affect onboarding messages

see https://github.com/18F/dolores-landingham-slack-bot/blob/develop/app/services/onboarding_message_employee_matcher.rb#L24 

we only send an onboarding message if there is no `SentMessage` record.

jmhooper:
I'm trying to remember why we need that uniqueness validation. Why do we consider multiple sent messages for an employee / message combination an invalid state?

jessieay:
We use these to determine if someone has already received an onboarding message. I think we can remove the validation *or* just make sure we do not create `SentMessage` records when testing messages

jmhooper:
This was fixed in #255

PR #255, opened 2017-03-10
Don't create SentMessage records for test messages

jmhooper:
Fixes issue(s) #250

- Don't create SentMessage records for test messages
- Add a feature spec to make sure sending a user a test message does not prevent them from receiving it as a scheduled onboarding message.

### DIFF
--- a/app/services/message_sender.rb
+++ b/app/services/message_sender.rb
@@ -1,9 +1,10 @@
 require "slack-ruby-client"
 
 class MessageSender
-  def initialize(employee, message)
+  def initialize(employee, message, test_message: false)
     @employee = employee
     @message = message
+    @test_message = test_message
   end
 
   def run
@@ -47,7 +48,7 @@ class MessageSender
 
   private
 
-  attr_reader :employee, :message
+  attr_reader :employee, :message, :test_message
 
   def client
     @client ||= Slack::Web::Client.new
@@ -62,14 +63,16 @@ class MessageSender
   end
 
   def create_sent_message(options)
-    SentMessage.create(
-      employee: options[:employee],
-      message: options[:message],
-      sent_on: Date.current,
-      sent_at: Time.current,
-      error_message: error_message(options[:error]),
-      message_body: formatted_message(options),
-    )
+    if !test_message
+      SentMessage.create(
+        employee: options[:employee],
+        message: options[:message],
+        sent_on: Date.current,
+        sent_at: Time.current,
+        error_message: error_message(options[:error]),
+        message_body: formatted_message(options),
+      )
+    end
   end
 
   def error_message(error)
--- a/spec/features/send_onboarding_message_after_test_message_spec.rb
+++ b/spec/features/send_onboarding_message_after_test_message_spec.rb
@@ -1,0 +1,36 @@
+require "rails_helper"
+
+feature "Send an oboarding message after a test message" do
+  scenario "Receiving a test message should not prevent a user from receiving the same scheduled onboarding message" do
+    days_after_start = 3
+    start_time = Time.parse("11:00:00 UTC")
+    employee_time_zone = "Eastern Time (US & Canada)"
+    current_time_for_employee = start_time.in_time_zone(employee_time_zone)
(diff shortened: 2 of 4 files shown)


jessieay:
Hey! protip:  if you leave a space between `#` and the issue number, github won't auto close the issue for you when this PR gets merged ;)

jmhooper:
Oops 🙃

app/services/message_sender.rb
```diff
@@ -1,9 +1,10 @@
 require "slack-ruby-client"
 
 class MessageSender
-  def initialize(employee, message)
+  def initialize(employee, message, options = {})
```
jessieay: thoughts on using a keyword arg that defaults to `nil` rather than an options hash? Passing the `test_message` arg would look exactly the same. What I like about using keyword args is that you do need to pass them with the correct key, whereas options hash lets you pass anything 

https://robots.thoughtbot.com/ruby-2-keyword-arguments

spec/features/send_onboarding_message_after_test_message_spec.rb
```diff
@@ -0,0 +1,43 @@
+require "rails_helper"
+
+feature "Send an oboarding message after a test message" do
+  scenario "Receiving a test message should not prevent a user from receiving the same scheduled onboarding message" do
```
jessieay: AWESOME spec description - so clear 🍔

spec/features/send_onboarding_message_after_test_message_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+    current_time_et_as_utc = Time.zone.utc_to_local(current_time_for_employee)
+    message_send_time = days_after_start.business_days.after start_time
+
+    onboarding_message = create(
+      :onboarding_message,
+      days_after_start: days_after_start,
+      time_of_day: current_time_et_as_utc
+    )
+    employee = create(:employee, started_on: start_time)
+
```
jessieay: extra newline here

spec/services/message_sender_spec.rb
```diff
@@ -25,6 +25,20 @@
     Timecop.return
   end
 
+  it "doesn't create a sent message if the test_message option is true" do
+    Timecop.freeze(Time.parse("10:00:00 UTC"))
```
jessieay: do we need timecop in this spec?

spec/features/send_onboarding_message_after_test_message_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+
+    Timecop.freeze(start_time) do
+      send_test_onboarding_message(employee, onboarding_message)
+      expect(employee_sent_messages(employee)).to be_empty
```
jessieay: I think we can eliminate this expectation since it is what the unit test is looking for (that test messages do not create a sent message record)

spec/features/send_onboarding_message_after_test_message_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+
+      Timecop.travel(days_after_start.business_days.after start_time)
+
+      OnboardingMessageSender.new.run
```
jessieay: let's make sure that this is the thing creating a `SentMessage` - so rather than the following line, we can have this:

```
expect { OnboardingMessageSender.new.run }.to change{ employee.sent_messages.count }.from(0).to(1)

app/services/message_sender.rb
jmhooper: I always forget about keyword args. Will do.

spec/services/message_sender_spec.rb
jmhooper: Nope, that's copy paste noise. Good catch

spec/features/send_onboarding_message_after_test_message_spec.rb
jmhooper: I.... didn't know you could do that. Whoa!

jmhooper:
Cleaned up and amended the commit

